PR #1, opened 2023-01-27
Connect server client

kdata12:
Connected the backend with the frontend. Created a button thats sends an HTTP GET to retrieve data from http://localhost:8000/api 

Test backend endpoints with Postman as well.

File changes:
In the backend, for dependencies, I added express. For dev-dependencies, I added cors, @type/cors, and ts-node-dev.

In the frontend, I just removed useless react boilerplate files.


frontend/src/App.tsx
```diff
@@ -1,24 +1,23 @@
-import React from 'react';
-import logo from './logo.svg';
-import './App.css';
+import axios from "axios";
+import { useState } from "react";
+const serverURL = 'http://localhost:8000'
 
 function App() {
+  const [message, setMessage] = useState("");
```
justinpchen94: nice use of hooks! this is optional but for typescript if you want to be extra safe with typings you can rewrite this line to be

```
const [message, setMessage] = useState<string>("");
```

additionally, this isn't super important for our use case but there's definitely room for debate as to whether you want to use `""` as the default message or something more explicit like `null` or `undefined`. the former keeps the types more consistent but the latter is preferred stylistically in some places to avoid situations like `""` not being a valid input but legally is a string

frontend/src/App.tsx
```diff
@@ -1,24 +1,23 @@
...
+  
+  const handleClick = async () => {
+    if (message) {
+      setMessage("")
+      return
+    }
```
justinpchen94: i know this code is mostly just demonstrative and not going to be here for very long but it's a good idea in general to make your code as self documenting as possible. it took me a second to realize that you wanted to use this code block to "reset" the message being shown and you could've abstracted this out into a fn called `resetMessage` or similar.

that being said, this code actually isn't super useful, since your `setMessage` call on line 15 would effectively change the message being shown. in this case, you are actually causing react to rerender the dependent child components twice, once to clear it to "", and then shortly followed by the message passed from the server. obviously in this case, the performance drop isn't noticeable but this stuff will make a difference once you're coding at scale!

frontend/src/App.tsx
```diff
@@ -1,24 +1,23 @@
...
-        <a
-          className="App-link"
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+      <h2>{message ? message : ""}</h2>
```
justinpchen94: no need to change but this can also be written as `message ?? ""`

this is a relatively recent addition to js and it's super helpful for these kinds of patterns
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

backend/package.json
```diff
@@ -4,14 +4,22 @@
...
+    "dev": "ts-node-dev index.ts",
+    "tsc": "tsc"
   },
   "author": "",
   "license": "ISC",
   "devDependencies": {
+    "@types/cors": "^2.8.13",
     "@types/express": "^4.17.16",
     "@types/node": "^18.11.18",
+    "cors": "^2.8.5",
```
justinpchen94: i think cors actually is needed in regular dependencies, not just devDependencies.

frontend/src/App.tsx
```diff
@@ -1,24 +1,23 @@
...
 function App() {
+  const [message, setMessage] = useState("");
+  
+  const handleClick = async () => {
+    if (message) {
+      setMessage("")
+      return
+    }
+    const res = await axios.get(`${serverURL}/api`)
+    const serverMessage = res.data.data
```
justinpchen94: not a necessity for this temporary code but it's good to think about how you would handle this route if it ever errored, since we are assuming in this code so far that the API always successfully returns

https://stackabuse.com/handling-errors-with-axios/

justinpchen94 (commented):
great first PR!

justinpchen94:
i also want to comment on the good PR body hygiene that you demonstrated. It's always a great thing to describe what your code is meant to do in the description body at the top. It shows empathy for your reviewers and also a consideration for all the future ppl that may have to reference this code again one day.

justinpchen94:
A great habit to get into that we'll keep for this project as well is to use the Task number in the title of this PR.

In cases where you only did one task, i'd do 
[T21] Connect the Server and Client

if there are more, i'd do the same title but add more details into the PR body with what tasks were handled (though again we're gonna try to keep it 1 task : 1 pr in the future) and links to the task in airtable

frontend/src/App.tsx
kdata12: Gotcha, I changed the hook to `useState<boolean>(false)`.

frontend/src/App.tsx
kdata12: I'm kind of confuse about why react would rerender the dependent chid component twice. I have the return statement after the setMessage(""). Can you elaborate on this? Also here's how I made the code more readable with a `resetMessage` fn:

      const handleClick = async () => {
      const displayingMessage = message ? true : false
       if (displayingMessage) {
            resetMessage()
            return 
      }
      ...

justinpchen94: oh sorry you're right! i glossed over it too quickly. having watched the video again, I realize now that the button operates as a reset button between API calls so this makes more sense now.

it looks good now!

frontend/src/App.tsx
kdata12: Clarification question: Do you want me to do `useState<boolean>(false)`? I find `useState<string>("")` to be simple and effective just for this temporary code.

frontend/src/App.tsx
kdata12: Fixed line 19 to `message ?? ""`. Pretty useful ngl.

frontend/src/App.tsx
justinpchen94: yeah you could've definitely used the ("") option. just wanted to point it out that it is something people give feedback on, depending on the style of the codebase

backend/package.json
kdata12: Ok, I moved it to dependencies.

frontend/src/App.tsx
kdata12: Ok, I added a try and catch block to catch network errors:

    try {
      const res = await axios.get(`${serverURL}/api`)
      const serverMessage = res.data.data
      setMessage(serverMessage)
    } catch (error) {
      //error has type 'unknown'. My fix: assert error as any.
      console.error("Error connecting to backend:", (error as any).message)
      alert(`Error connecting to backend: ${(error as any).message}`)
    }

frontend/src/App.tsx
```diff
@@ -1,24 +1,35 @@
...
 
 function App() {
+  const [message, setMessage] = useState<string>("");
+
+  const resetMessage = () => {
+    setMessage("")
+  }
+
+  const handleClick = async () => {
+    const displayingMessage = message ? true : false
```
justinpchen94: nit: let's use a _slightly_ more descriptive variable name here instead of `displayingMessage` and do `isDisplayingMessage`

in general it's good practice to prefix boolean types with `is` or `are` to be a little clearer

https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig

justinpchen94: also you don't have to make this change but this can be written as

`const displayingMessage = !!message`

https://www.codingem.com/javascript-double-exclamation-operator/

justinpchen94 (approved):

PR #3, opened 2023-02-04
T48 and T49: Logging Library + Environment Variables

nathan0522:
- Added Morgan as logging library
- Incorporating dotenv to use environment variables
- Created sample-env


justinpchen94 (approved):
lgtm!

PR #7, opened 2023-02-05
T44: Create users table

jessicahe004:
added user table: primary ID (auto-increment), firstname, lastname, email (unique), password 

deleted commented previous tables as well


backend/prisma/migrations/20230205202717_create_user_table/migration.sql
```diff
@@ -0,0 +1,14 @@
...
+    `id` INTEGER NOT NULL AUTO_INCREMENT,
+    `email` VARCHAR(191) NOT NULL,
+    `firstName` VARCHAR(191) NOT NULL,
+    `lastName` VARCHAR(191) NOT NULL,
+    `birthdate` DATETIME(3) NOT NULL,
+
+    UNIQUE INDEX `User_email_key`(`email`),
+    UNIQUE INDEX `User_firstName_key`(`firstName`),
+    UNIQUE INDEX `User_lastName_key`(`lastName`),
+    UNIQUE INDEX `User_birthdate_key`(`birthdate`),
```
chinkevin25: I don't think we're going to want to include a `UNIQUE INDEX` for these three fields. 

Otherwise there we won't be able to accommodate having two Kevins, or Two Justins

backend/prisma/schema.prisma
```diff
@@ -7,28 +7,10 @@ datasource db {
...
-//   name    String?
-//   posts   Post[]
-//   profile Profile?
-// }
+model User {
+  id      Int      @id @default(autoincrement())
+  email   String   @unique
+  firstName    String @unique
+  lastName String @unique 
+  birthdate DateTime @unique
```
chinkevin25: same here, these three probably shouldn't be unique

chinkevin25:
Also, could you fill out the PR description

backend/prisma/migrations/20230205202717_create_user_table/migration.sql
justinpchen94: yeah you don't have to update this file, just update the schema.prisma file and then redo the prisma generate command

backend/prisma/schema.prisma
justinpchen94: also it should be `birthDate`

justinpchen94:
let's put a task number into the title too please!

backend/prisma/schema.prisma
```diff
@@ -7,28 +7,12 @@ datasource db {
...
-//   user   User    @relation(fields: [userId], references: [id])
-//   userId Int     @unique
-// }
+model User {
+  id      Int      @id @default(autoincrement())
+  email   String   @unique
+  firstName    String 
+  lastName String 
+  passwords String 
+  birthdate DateTime
```
chinkevin25: as justin pointed out, this should be `birthDate` (missed it the first time my bad)

backend/prisma/schema.prisma
```diff
@@ -7,28 +7,12 @@ datasource db {
...
-//   user   User    @relation(fields: [userId], references: [id])
-//   userId Int     @unique
-// }
+model User {
+  id      Int      @id @default(autoincrement())
+  email   String   @unique
+  firstName    String 
+  lastName String 
+  passwords String 
+  birthDate DateTime
```
justinpchen94: we might need to add an option to make this nullable because we aren't gonna ask birthday on the signup screen I think

backend/prisma/schema.prisma
```diff
@@ -7,28 +7,12 @@ datasource db {
...
-// }
-
-// model Profile {
-//   id     Int     @id @default(autoincrement())
-//   bio    String?
-//   user   User    @relation(fields: [userId], references: [id])
-//   userId Int     @unique
-// }
+model User {
+  id      Int      @id @default(autoincrement())
```
chinkevin25: No change necessary but wanted to have a conversation about this for us to think about.

So the type `Int` in figma is the MySQL INT  has a max size of 4,294,967,295. So what would happen if you were to register the 4,294,967,296th user?

It's not really something we need to worry about in this project, (if we got 4 billion users, we should all retire and buy a yacht) but it is something to think about when we make choices like this.

Another consideration to talk about is sequentially incremented ids can lead to some very interesting data vulnerabilities.

For example, lets imagine there is a profile page that has the route of 

`weWrap.com/users/1/profile`

where `1` the id of is Kevin Chin's `User` Row

If I were curious enough to, I might try to directly access

`weWrap.com/users/2/profile` where 2 is Justin's `User` Row

Because they are sequential, I'm about to guess and try to find other peoples profiles, and there is improper security, i might be able to see sensitive data that was not mine.

For reference I think Capitol One had a data leak once because someone found there id number, and then just incremented it by 1 and found other peoples Credit Card numbers.


There are alternative ways of using unique ids (uuids) https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#uuid

This way both overcomes the sequential id numbers as well as increases the maximum possible combinations to nearly infinite.

backend/prisma/schema.prisma
```diff
@@ -7,28 +7,12 @@ datasource db {
...
+  email   String   @unique
+  firstName    String 
+  lastName String 
+  passwords String
```
chinkevin25: A user probably will only have one password here, so this field is probably just `password`

backend/prisma/schema.prisma
```diff
@@ -7,28 +7,12 @@ datasource db {
...
-//   userId Int     @unique
-// }
+model User {
+  id      Int      @id @default(autoincrement())
+  email   String   @unique
+  firstName    String 
+  lastName String 
+  passwords String 
+  birthDate DateTime 
+}
```
chinkevin25, line 17: We don't need to do this for this project but I do want to bring up that sometimes we might want to consider splitting up our schema files into separate folders.

As schemas get really large, we might want to move them into different folders to make it easier to find.

i.e. this schema file might want to be `backend/prisma/user/schema.prisma`

In our project, the schema probably won't be too complicated that we don't need to split it up. Just something to think about

chinkevin25 (changes requested):

backend/prisma/schema.prisma
jessicahe004: okay i fixed the passwords -> password

backend/prisma/schema.prisma
jessicahe004: i will change id -> uuid to prevent this security issue

backend/prisma/schema.prisma
jessicahe004: okay i will change birthdate -> birthDate

backend/prisma/schema.prisma
jessicahe004: okay i will make this nullable

justinpchen94 (approved):
looking good!

backend/prisma/schema.prisma
```diff
@@ -7,28 +7,12 @@ datasource db {
...
-// }
-
-// model Profile {
-//   id     Int     @id @default(autoincrement())
-//   bio    String?
-//   user   User    @relation(fields: [userId], references: [id])
-//   userId Int     @unique
-// }
+model User {
+  id   String @id @default(cuid())
```
justinpchen94: also, im curious if you had a specific reason for choosing `cuid` here

chinkevin25 (approved):
Looking good!

fahadakhan2 (approved):
Solid work!

backend/prisma/schema.prisma
jessicahe004: i confused cuid with uuid, but ultimately looking into the differences i decided that uuid was the more secure choice. this is because when i read up on the cuid documentation, there was a deprecation. also prisma supports uuid from its theory whereas cuid support is dependent on external code.

jessicahe004:
revised users table and got the green light from justin, fahad, and kevin

PR #8, opened 2023-02-05
Update README.md

jessicahe004:
update instructions regarding database migrations


justinpchen94 (commented):
let's also add in those extra steps we talked about in the channel around the dotenv cli and prefixing the migrate command with the .env path!

let's add it to the server/prisma/README file

jessicahe004:
Leaving the README.md file be.

PR #9, opened 2023-02-06
T61: edit Prisma/readme.md

jessicahe004:
provided instructions for db migration

### DIFF
--- a/backend/prisma/README.md
+++ b/backend/prisma/README.md
@@ -4,3 +4,10 @@ Replace the database connection string in the schema.prisma with your own.
 
 Then uncomment out the rest of the code starting from line 10.
 
+DB Migration: https://www.prisma.io/docs/concepts/components/prisma-migrate/get-started
+- *to access the correct .env file use: dotenv -e ../.env
+- 1. make sure to install the dotenv dependency: npm install -g dotenv-cli 
+- 2. perform DB migration: dotenv -e ../.env npx prisma migrate dev --name init (replace init with relevant title) 
+
+
+
--- a/backend/prisma/README.md
+++ b/backend/prisma/README.md
@@ -1,12 +1,10 @@
-https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-typescript-postgres
+Prefix all commands to access the correct .env file with: dotenv -e ../.env
 
-Replace the database connection string in the schema.prisma with your own.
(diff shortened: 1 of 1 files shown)


backend/prisma/README.md
```diff
@@ -4,3 +4,10 @@ Replace the database connection string in the schema.prisma with your own.
 
 Then uncomment out the rest of the code starting from line 10.
 
+DB Migration: https://www.prisma.io/docs/concepts/components/prisma-migrate/get-started
+- *to access the correct .env file use: dotenv -e ../.env
```
justinpchen94: nit: let's add something along the lines of `prefix your command with: dotnev...` since this applies to any prisma command, not just the one below

jessicahe004: okay i will clarify the prisma command

backend/prisma/README.md
```diff
@@ -1,6 +1,11 @@
-https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-typescript-postgres
+Prefix all commands to access the correct .env file with: dotenv -e ../.env
+
+Connecting DB: https://www.prisma.io/docs/getting-started/setup-prisma/start-from-scratch/relational-databases/connect-your-database-typescript-postgres
+- 1. Replace the database connection string in the schema.prisma with your own.
```
justinpchen94: this should actually be replaced in side the `.env` file instead!

backend/prisma/README.md
```diff
@@ -1,6 +1,11 @@
...
+
+DB Migration: https://www.prisma.io/docs/concepts/components/prisma-migrate/get-started
+- 1. make sure to install the dotenv dependency: cd backend npm install -g dotenv-cli
```
justinpchen94: fyi: the `-g` flag here means global, so it actually doesn't matter what directory you're in when you run this command! so we can remove the `cd backend` step

justinpchen94:
let's also make a new task in the tracker to track this piece of work with, so we can put a task number in the title!

backend/prisma/README.md
jessicahe004: okay i will revise that

backend/prisma/README.md
jessicahe004: okay i will rrevise that too

justinpchen94 (approved):

PR #10, opened 2023-02-10
T58: Add Google OAuth strategy from passport + File restructing

kdata12:
**Implemented google OAuth using**: https://www.passportjs.org/packages/passport-google-oauth20/

Added GET /auth/google endpoint that will initialize the google auth process to the backend. 

Added logic to find or create new user based on whether user has googleID or not. 

Added serialization process that attaches the user id to a cookie

**File restructuring:**

1. Moved all the routes to a route folder

2. Only have index.ts connect our app to the local server

3. Created a new file called app.ts that will have all of the dependencies, middleware and routing.

4. Created a utils folder to store custom middleware and other services

5. Moved prisma to Utils/prismaClient to prevent repetitive importing and initializing PrismaClient

No changes to frontend.

Google OAuth diagram:
https://drive.google.com/file/d/1sNxGNtIp7yJ4hLmqylowID2ojrH5kjZR/view?usp=share_link

Google Strategy explanation:
https://drive.google.com/file/d/1d-HPfTsI3hqn9bzadmWQGN7AiqrtA3n3/view?usp=share_link


backend/app.ts
```diff
@@ -0,0 +1,76 @@
...
+import prisma from './utils/prismaClient';
+
+const app = express();
+
+dotenv.config();
+app.use(morgan("dev"));
+app.use(cors());
+app.use(express.json());
+app.use(session({
+    secret: "secretcode",
```
justinpchen94: should this be plaintext like this or also saved as a secret somewhere?

backend/app.ts
```diff
@@ -0,0 +1,76 @@
+import * as dotenv from 'dotenv';
+import path from 'path';
+dotenv.config({ path: path.resolve(__dirname, "../.env") });
+import express from 'express';
+import cors from 'cors';
+import morgan from 'morgan';
+import session from 'express-session';
+import passport from 'passport';
+const GoogleStrategy = require('passport-google-oauth20').Strategy;
```
chinkevin25: Do you think we can keep this as an `import` so we are staying consistent with the methodolgy?

backend/app.ts
chinkevin25: We want to make sure this is secret, so technically, this should be pulled in from an `ENV var

backend/app.ts
```diff
@@ -0,0 +1,76 @@
...
+const app = express();
+
+dotenv.config();
+app.use(morgan("dev"));
+app.use(cors());
+app.use(express.json());
+app.use(session({
+    secret: "secretcode",
+    resave: true,
+    saveUninitialized: true,
```
justinpchen94: genuine question, maybe good for @chinkevin25 to chime in on too: should these options both be set to true?

from docs: https://www.npmjs.com/package/express-session

resave:

> How do I know if this is necessary for my store? The best way to know is to check with your store if it implements the touch method. If it does, then you can safely set resave: false. If it does not implement the touch method and your store sets an expiration date on stored sessions, then you likely need resave: true.

saveUninitialized:
> Forces a session that is "uninitialized" to be saved to the store. A session is uninitialized when it is new but not modified. Choosing false is useful for implementing login sessions, reducing server storage usage, or complying with laws that require permission before setting a cookie. Choosing false will also help with race conditions where a client makes multiple parallel requests without a session.

backend/index.ts
```diff
@@ -1,37 +1,12 @@
...
-    res.status(200).json({
-        data: "Together we wrap!"
-    })
-})
+const server = http.createServer(app)
 
 const port = 8000;
 
-app.listen(8000, () => {
+server.listen(port, () => {
```
justinpchen94, line 10: ❤️

backend/package.json
```diff
@@ -7,16 +7,20 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "dev": "ts-node-dev index.ts",
     "tsc": "tsc",
-    "postinstall": "prisma generate"
+    "postinstall": "prisma generate",
+    "prismaMigrate": "npx prisma migrate dev --name"
```
justinpchen94, line 11: does this command work as is without having to provide a path to the .env file? i know some other members had issues with this

backend/routes/google-oauth.ts
```diff
@@ -0,0 +1,17 @@
+import express from "express";
```
justinpchen94: nit: let's be consistent with file naming, since it looks like we use camel case throughout, let's rename this file to be `/routes/googleOAuth` 

that being said, this file in particular will likely contain future authentication related routes so it's up to you whether you want to keep this file solely for google oauth or rename to be more generic for all auth

backend/routes/test-route.ts
```diff
@@ -0,0 +1,21 @@
+import express from "express";
```
justinpchen94: nit: consistent casing
`/routes/testRoute.ts`

backend/prisma/schema.prisma
```diff
@@ -8,11 +8,11 @@ datasource db {
...
 
 model User {
-  id   String @id @default(uuid())
-  email   String   @unique
-  firstName    String 
-  lastName String 
-  password String 
-  birthDate DateTime? 
+  id        String    @id @default(uuid())
+  email     String?    @unique
```
justinpchen94: unfortunately since the prev migration that you deleted has been run by other members, we will begin running into migration issues. however, given that it is early in the project, we have 2 options

1) hacky. have everyone drop their db's and run a new migration that creates the User table based on the newly modified schema you have made here

2) best practice. write a new migration that handles altering the table schema to transition from the old User table schema to this new one. i _believe_ prisma will handle most of this logic for you but you'll need to figure out how that part works (and hopefully pass the knowledge on to your teammates!)

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules/
 dist/
-.env
\ No newline at end of file
+.env
+backend/prisma/migrations/
```
justinpchen94: we probably shouldn't be ignoring all migrations! these will need to be run by everyone in order to keep everyone's dbs up to date

justinpchen94 (commented):
deferring the auth specific code to @chinkevin25 but left some comments on the other parts

justinpchen94:
also it'd be great to attach an image or pdf of your auth diagram to this pr, would help ppl immensely!

backend/app.ts
kdata12: Oopsie, I will move it to .env

backend/app.ts
kdata12: @chinkevin25 can you take a look at two options? The way I interpreted **resave** is that when a user visits a page, their session will saved even when the session was not modified. Always keeping the user session on our database.

For ** saveUninitialized**, as soon the user visits the website, a session is created for them. I think this should be true (?) unless we only create sessions for people with accounts.

backend/package.json
kdata12: Yeah, it does work for me without providing a path to .env. I don't quite understand by what you mean by having to provide a path to the .env file, can you elaborate on this?

backend/routes/google-oauth.ts
kdata12: Changed file name to googleOAuth.

I think I will keep it only for google. I've thought it about it, when we implement other auth services, I think it will be easier to find with separate files.

backend/prisma/schema.prisma
kdata12: Ok, I will research more into schema migration from old to new.

kdata12: Fortunately, I altered column **email**, **password**, **birthDate** to _nullable_ and added _nullable_ column **googleID**, these changes to the schema will not cause our database to reset.

After researching, I found a method to migrate schema without data loss. When prompted a database reset, we can use the **prisma migrate --create-only** command to create a .sql file with the new schema from schema.prisma file, this will only create a new migration but **does not apply the changes**. Then, we can edit the .sql file to alter columns: drop column, set default values, set non-null constraints etc. Finally, after making the necessary changes to the .sql file, we can just use **prisma migrate dev** to apply the custom migration to our database. As long as our database end state matches with schema.prisma, we can migrate without data loss.

Prisma doc:
https://www.prisma.io/docs/reference/api-reference/command-reference#options-6

backend/app.ts
chinkevin25: For `saveUninitialized`, like you mentioned it'll save a session to the session store immediately. Like you mentioned it'll be true if we needed to store information on the session other than login. We might include this if we were doing more tracking. For now, however, we're only going to be doing login sessions. So we could opt for false, if we wanted to. Either `true` or `false` works for our purpose, so it's just best to understand when you'd want it on.

Another reason we might want immediate session saving, is if we needed to carry information from a logged out state to a logged in state, or if we wanted to see if "repeat" clients came to visit. I'd lean to leaving it `false` until we have a need to turn it on just to save DB space.

chinkevin25: For resave, since you are not setting a `store` value, I believe you're using the default MemoryStorage which I don't think has `touch` implemented. So you should have resave `true` as you have it.

It's worth mentioning that `MemoryStore` will need to be replaced in production, but can be done later.

backend/prisma/schema.prisma
chinkevin25: Do you know if we will run into problems in our application if we don't have an email provided for a user? 

When we thought about local strategy, it would require everyone to have an email associated w/ weWrap.

Is there a way to pull their email and store it in our database from the oauth response?

kdata12: Yup, I figured out how to add the email scope when we request user data. I will change our email attribute back to non-nullable.

kdata12: Ok, I will change `resave` to true and `saveUninitialized` to false to save DB space for now.

kdata12:
After pulling and merging with the main branch, I had to make a few changes to the file:

- I created a file in ./backend/routes/loginAuth.ts file to store all of FK's login logic with an express router object.

- Added salt column as optional, kept password column as not required

- I think I messed up some merge conflicts in the package-lock.json file so I had to delete node_module and package-lock.json with `rm -rf node_module/` and `rm package-lock.json` and reinstall everything with `npm i`

backend/app.ts
```diff
@@ -0,0 +1,87 @@
...
+const db = prisma;
+const app = express();
+
+app.use(morgan("dev"));
+app.use(cors());
+app.use(express.json());
+
+const secretcode = process.env.SESSION_SECRET as string;
+const clientID = process.env.GOOGLE_CLIENT_ID as string;
+const clientSecret = process.env.GOOGLE_CLIENT_SECRET as string;
```
chinkevin25, line 24: no change needed but it's common to put all of your ENV vars in a `config.ts` file that gets imported to files as you need it.

chinkevin25 (approved):

PR #11, opened 2023-02-11
[T41] Added backend logic for logging in a user

fahadakhan2:
**Server Side: index.ts**
*  A GET request to the route /login/password renders a login form
*  A POST request to the same route will attempt to authenticate the user with the provided username and password.
* The code uses the passport library to authenticate the user, which is set up with the LocalStrategy.
* The mysql library is used to connect to the database and retrieve the user information.
* Then the strategy uses the crypto library to hash the password and compare it to the hashed password stored in a database. 
* If the comparison is successful, the user is considered authenticated. If not, the function returns an error indicating that the username or password is incorrect.


backend/index.ts
```diff
@@ -6,6 +6,7 @@ import cors from 'cors';
 import { PrismaClient } from '@prisma/client'
 import morgan from 'morgan';
 
+
```
justinpchen94: let's rm - generally unspoken rule that 1 blank line is sufficient

backend/index.ts
```diff
@@ -35,3 +36,53 @@ const port = 8000;
 app.listen(8000, () => {
     console.log(`Server started: http://localhost:${port}/`);
 }); 
+
+
+
+// Configure Strategy for Login
+
+// user is prompted to sign in with their username and password by rendering a form
+// defining a route
```
justinpchen94: love that you use comments to guide your work, but I don't think these are necessary to keep here!

backend/index.ts
```diff
@@ -35,3 +36,53 @@ const port = 8000;
...
+    res.status(200).send("we logged in")
+  });
+
+var passport = require('passport');
+var LocalStrategy = require('passport-local');
+var crypto = require('crypto');
+var mysql = require('mysql12')
+
+// connecting to the database
+var connection = mysql.createConnection({
```
justinpchen94: this will be caught by our linter eventually but modern js uses `let` and `const` and not `var` anymore. let's also convert these into `import` statements instead of `require`'s

backend/index.ts
```diff
@@ -35,3 +36,53 @@ const port = 8000;
...
+    host: 'localhost',
+    user: 'root',
+    password: '',
+    database: 'wewrap_local'
```
justinpchen94: whoever gets to it first, you or @jesshe004, will want to move these fields into the `.env`!

backend/index.ts
```diff
@@ -35,3 +36,53 @@ const port = 8000;
...
+  });
+  
+// checking for user info in the database
+passport.use(new LocalStrategy(function verify(username:any, password:any, cb:any) {
```
justinpchen94: will leave this block to @chinkevin25 to double check

backend/prisma/schema.prisma
```diff
@@ -8,11 +8,10 @@ datasource db {
 }
 
 model User {
-  id   String @id @default(uuid())
-  email   String   @unique
-  firstName    String 
-  lastName String 
-  password String 
-  birthDate DateTime? 
+  id        String    @id @default(uuid())
```
justinpchen94, line 11: ty for formatting update!

frontend/package.json
```diff
@@ -10,7 +10,7 @@
     "@types/node": "^16.18.11",
     "@types/react": "^18.0.27",
     "@types/react-dom": "^18.0.10",
-    "axios": "^1.2.5",
+    "axios": "^1.3.2",
```
justinpchen94: your frontend package.lock diff looks particularly large for such a small change - can you try deleting the package-lock and re-generating it? (just `npm i`)

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react';
+import axios from 'axios';
+
+
```
justinpchen94: let's rm

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+          axios.post("http://localhost:8000/login/password", {
+            username,
+            password,
+          })
+          .then(response => {
+            console.log(response.data);
+          })
+          .catch(error => {
+            console.error(error);
+          });
```
justinpchen94: i think this works just fine! but let's take a stab at converting this into async/await style js rather than promises

https://rapidapi.com/guides/axios-async-await

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+      };  
+    
+    
+    return (
+        <form action="/login/password" method="post" onSubmit={handleSubmit}>
+            <div>
+                <label htmlFor="username">Username</label>
+                <input id="username" name="username" type="text" autoComplete="username" required value={username} onChange={event => setUsername(event.target.value)} />
```
justinpchen94: i know the onChange here is quite small but it's good practice to try and make the return of the component as structural-only as possible and pull any functions and business logic into separate variables above. so let's pull this onChange out and define it separately!

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,40 @@
...
+            </div>
+            <div>
+                <label htmlFor="current-password">Password</label>
+                <input id="current-password" name="password" type="password" autoComplete="current-password" required value={password} onChange={event => setPassword(event.target.value)} />
```
justinpchen94: same with this onchange

justinpchen94 (commented):
good looking first pass! left some comments

backend/index.ts
fahadakhan2: Will remove rn!

backend/index.ts
fahadakhan2: Okay good to know will import these instead.

backend/index.ts
fahadakhan2: I'm a little bit confused as to how to move the list of these variables into .env and then properly relate back to them on the index.ts file when creating the mysql connection. I'd be willing to do this if Jess doesn't get to it first but I just need some clarification.

justinpchen94: here it will look like:
```
var connection = mysql.createConnection({
    host: process.env.DB_HOST,
    user: process.env.DB_USER,
    password: process.env.DB_PASSWORD,
    database: process.env.DB_NAME
  });

```

and in the .env

```
DB_HOST="localhost"
DB_USER="root"
DB_PASSWORD=""
DB_NAME="wewrap_local"

```

fahadakhan2: Oooo this really helps understanding it thanks!

frontend/package.json
fahadakhan2: Fixed!

frontend/src/Components/loginForm.tsx
fahadakhan2: Okay fixed! By the reading its to make the code look more clean right?

frontend/src/Components/login.css
```diff
@@ -0,0 +1,69 @@
...
+
+
+
+
+
+
+
+
+
+
```
justinpchen94: nit: only 1 new line at the bottom!

frontend/src/App.tsx
```diff
@@ -29,7 +30,8 @@ function App() {
   return (
     <div className="App">
       <h2>{message ?? ""}</h2>
-      <button onClick={handleClick}>click me!</button>
+      <LoginForm />
+      {/* <button onClick={handleClick}>click me!</button> */}
```
justinpchen94: let's either uncomment or delete this line!

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,54 @@
...
+        <section>
+          <img src='./logo192.png' alt='placeholder for wewrap logo'></img>
+          <h1>Log in to your account</h1>
+          <p className='subheader'>Don't have an account? Sign up</p>
+
+          <form action="/login/password" method="post" onSubmit={handleSubmit}>
+              <div>
+                  <p className='form_text'>username*</p>
+                  <label htmlFor="username"></label>
+                  <input id="username" name="username" type="text" autoComplete="username" required value={username} onChange={event => handleUsernameChange(event, setUsername)}/>
```
justinpchen94: not sure if this is jsx specific but `autoComplete` I think only accepts `on` or `off` as values according to this: https://www.w3schools.com/tags/att_input_autocomplete.asp

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,54 @@
...
+              </div>
+              <div>
+                  <p className='form_text'>password*</p>
+                  <label htmlFor="current-password"></label>
+                  <input id="current-password" name="password" type="password" autoComplete="current-password" required value={password} onChange={event => handlePasswordChange(event, setPassword)}/>
```
justinpchen94: same autocomplete comment

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,54 @@
...
+                  <p className='form_text'>remember me</p>
```
justinpchen94: nit: we use `-` instead of `_` in other places, so let's be consistent with choosing one or the other

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,54 @@
...
+    setUsername(event.target.value);
+  };
+
+  const handlePasswordChange = (event: React.ChangeEvent<HTMLInputElement>, setPassword: React.Dispatch<React.SetStateAction<string>>) => {
+    setPassword(event.target.value);
+  };
+
+    return (
+        <section>
+          <img src='./logo192.png' alt='placeholder for wewrap logo'></img>
```
justinpchen94: depending on the timing of when this image will get provided, it's worth putting a comment here and then creating an entry in the task tracker to make sure we don't forget to complete this! it's very common in industry to have a comment that reads like:

// TODO (justin) T99: refactor this code after ____

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,54 @@
...
+        } catch(error) {
+            console.error(error);
+        }
+    };
+    
+  const handleUsernameChange = (event: React.ChangeEvent<HTMLInputElement>, setUsername: React.Dispatch<React.SetStateAction<string>>) => {
+    setUsername(event.target.value);
+  };
+
+  const handlePasswordChange = (event: React.ChangeEvent<HTMLInputElement>, setPassword: React.Dispatch<React.SetStateAction<string>>) => {
```
justinpchen94: so you actually don't need to pass `setPassword` in as an argument here since `setPassword` is within the scope of the entire component and you can just access it implicitly

frontend/src/Components/loginForm.tsx
justinpchen94: also related to my other comment but once you change the fn signature for `handlePasswordChange` to just requiring an event, then you can change this to be `onChange={handlePasswordChange}`

frontend/src/Components/loginForm.tsx
justinpchen94: same comment here with `handleUsernameChange` and the `setUsername` argument change

`onChange={handleUsernameChange}`

justinpchen94 (commented):
looking better, couple more nitpicky comments!

frontend/src/Components/loginForm.tsx
fahadakhan2: Okay makes sense.

frontend/src/Components/loginForm.tsx
fahadakhan2: Done!

frontend/src/Components/loginForm.tsx
fahadakhan2: Okay will change it to underscores.

frontend/src/Components/loginForm.tsx
fahadakhan2: Changed the autoComplete to "on"

frontend/src/Components/loginForm.tsx
fahadakhan2: Done!

backend/index.ts
chinkevin25: we try to avoid using `any` as a type where we can. Do you think we can make our types more specific?

backend/index.ts
```diff
@@ -35,3 +40,38 @@ const port = 8000;
...
+
+const connection = mysql.createConnection({
+    host: process.env.DB_HOST,
+    user: process.env.DB_USER,
+    password: process.env.DB_PASSWORD,
+    database: process.env.DB_NAME
+  });
+  
+passport.use(new LocalStrategy(function verify(username:any, password:any, cb:any) {
+    connection.query('SELECT * FROM users WHERE username = ?', [ username ], function(err:any, user:any) {
```
chinkevin25: So this is the way we would do it if we weren't using Prisma. Because we're using Prisma, we should be important Prisma models and using them instead.

backend/index.ts
```diff
@@ -35,3 +40,38 @@ const port = 8000;
...
+      if (err) { return cb(err); }
+      if (!user) { return cb(null, false, { message: 'Incorrect username or password.' }); }
+  
+      crypto.pbkdf2(password, user.salt, 310000, 32, 'sha256', function(err:any, hashedPassword:any) {
+        if (err) { return cb(err); }
+        if (!crypto.timingSafeEqual(user.hashed_password, hashedPassword)) {
+          return cb(null, false, { message: 'Incorrect username or password.' });
+        }
+        return cb(null, user);
+      });
```
chinkevin25: This is a little hard to read for me. We already have the verify function named so maybe we can break it up so that it's a little easier to read?

backend/index.ts
chinkevin25: also, can we move all of our auth middleware logic into its own folder so we can isolate the index from this login specific logic

backend/index.ts
```diff
@@ -35,3 +40,38 @@ const port = 8000;
...
+    console.log(req.body)
+    res.status(200).send("we logged in")
+  });
+
+const connection = mysql.createConnection({
+    host: process.env.DB_HOST,
+    user: process.env.DB_USER,
+    password: process.env.DB_PASSWORD,
+    database: process.env.DB_NAME
+  });
```
chinkevin25: We should also have a Prisma way of doing a connection instead of doing it like this.

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,55 @@
...
+        <section>
+          {/* TODO (Fahad) T78: refactor this code after wewrap logo can be downloaded */}
+          <img src='./logo192.png' alt='placeholder for wewrap logo'></img>
+          <h1>Log in to your account</h1>
+          <p className='subheader'>Don't have an account? Sign up</p>
+
+          <form action="/login/password" method="post" onSubmit={handleSubmit}>
+              <div>
+                  <p className='form_text'>username*</p>
+                  <label htmlFor="username"></label>
```
chinkevin25: We should use the `label` to provide the actual label for the input rather than a `p` tag.

chinkevin25: https://stackoverflow.com/questions/11197671/use-css-to-automatically-add-required-field-asterisk-to-form-inputs

also a cool way to do this as well. You'll want to also probably create some CSS to make the label format the way you want it.

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,55 @@
...
+                  <input id="username" name="username" type="text" autoComplete="on" required value={username} onChange={handleUsernameChange}/>
+              </div>
+              <div>
+                  <p className='form_text'>password*</p>
+                  <label htmlFor="current_password"></label>
```
chinkevin25: same deal here

backend/index.ts
fahadakhan2: Done!

frontend/src/Components/loginForm.tsx
fahadakhan2: Fixed!

justinpchen94 (approved):

PR #13, opened 2023-02-14
[T40] Build Login Page Frontend

fahadakhan2:
**loginForm.tsx & login.css**
* Used the axios library to make a POST request to a backend API at http://localhost:8000/login/password with the email and password as the payload of the request.
* When the form is submitted, the handleSubmit function is triggered
* An API request is made using axios.post() and logs the response data to the console if the request is successful. If the request is not successful, it logs the error to the console.
* The component returns a form that has two input fields for the email and password, and a submit button. The form uses the onSubmit prop to attach the handleSubmit function to the submit event.
* CSS styling applied to login page


frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const [password, setPassword] = useState<string>('');
+    
+    const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+        event.preventDefault();
+        try {
+          const response = await axios.post("http://localhost:8000/login/password", {
+            email,
+            password,
+          });
+          console.log(response.data)
```
chinkevin25, line 16: No change just yet but thinking in the future when we get a successful response, what should happen? 

We may want to add a TODO to document that

frontend/src/Components/login.css
```diff
@@ -0,0 +1,138 @@
...
+  flex-direction: column;
+  font-size: 14px;
+  margin-bottom: 0%;
+}
+
+.email{
+  margin-bottom: 8px;
+}
+
+input{
```
chinkevin25: I'm purposefully going to ask us not to change this knowing that this might be problematic. This CSS rule will format all inputs, so it's possible this will mess w/ the sign up pages styling too.

frontend/src/App.tsx
```diff
@@ -1,35 +1,10 @@
...
-      //error has type 'unknown'. My fix: assert error as any.
-      console.error("Error connecting to backend:", (error as any).message)
-      alert(`Error connecting to backend: ${(error as any).message}`)
-    }
-  }
   return (
     <div className="App">
-      <h2>{message ?? ""}</h2>
-      <button onClick={handleClick}>click me!</button>
+      <LoginForm />
```
chinkevin25, line 7: Something to think about because it'll probably cause a merge conflict with Jess anyways is.

How do we determine which page should show on which URL? Hint: React Router is the package we'll probably be using.

frontend/src/Components/loginForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+              <div className='user_helper'>
+                <label className='remember_me_and_checkbox'>
+                  <input type="checkbox" name="remember_me" className="remember_me"/>Remember Me
+                </label>
+                {/* T81 (FK) Create and add route to forgot password page once that page is built */}
+                <p className='forget_password'><a href=''>Forgot password?</a> </p>
+              </div>
+              <div>
+                  <button className='login_button' type="submit">Log in</button>
+              </div>
```
chinkevin25, line 55: This button right now can be pressed with nothing filled out. What could we do to prevent this?

chinkevin25 (changes requested):

frontend/src/Components/login.css
fahadakhan2: Okay but how would we get the input boxes to be the shape that Ray envisioned on figma?

frontend/src/Components/loginForm.tsx
fahadakhan2: Would it be that we redirect the user to a new page. Like maybe the landing page, their profile, or someplace else?

frontend/src/Components/loginForm.tsx
fahadakhan2: I thought I prevented that when the user tries clicking login without email/passoword being filled it it tells them to fill out the fields. I did this by including 'required for the inputs of email and pass.

frontend/src/App.tsx
fahadakhan2: I just did some reading on React Router and I get the concept of it but not sure where I should start to avoid the merge conflicts.

frontend/src/Components/loginForm.tsx
chinkevin25: Yup! Exactly, I don't know if we have a page in mind yet so it might be redirect to whatever page they were trying to access as well. 

I.e. trying to access their friend list would redirect them to login and then redirect back to the friend list page.

frontend/src/Components/login.css
chinkevin25: right, so this is also a TODO for @jesshe004 to install and configure css-modules! That will fix this issue.

frontend/src/App.tsx
chinkevin25: No need to avoid the conflicts, just when your PR merges or Jess's be ready to resolve conflicts!

frontend/src/Components/loginForm.tsx
chinkevin25: Ooh maybe it does do this! I didn't actually check myself -- does it prevent you from clicking the button?

frontend/src/Components/login.css
justinpchen94: yeah for now though you can simply change this `input` selector into a new class, and then apply that class to each of the inputs on your login page

frontend/src/Components/loginForm.tsx
justinpchen94: yeah +1 to kevin, let's add a todo here at the very least to document that we still need to fix this

frontend/src/Components/loginForm.tsx
justinpchen94: yeah im actually curious too, i've never really played with the "required" option before

justinpchen94 (commented):
generally looks good enough to merge after you reply to kevins comments!

frontend/src/Components/loginForm.tsx
fahadakhan2: Added!

frontend/src/Components/login.css
fahadakhan2: Okay used their class names instead for styling.

frontend/src/Components/loginForm.tsx
fahadakhan2: yeah it does it redirects you to fill out the fields. I will send you a ss on discord.

chinkevin25 (approved):

PR #14, opened 2023-02-15
T47: Add backend logic for signing up a user

jessicahe004:
Handles the axios post request from /signup for the registration of new users. 
- creates random salt and also hashes client password. 
- inserts email, first name, last name, hashed password, and salt into the database. 

*still need to handle general errors: unique email error & potential db being down & or website crashing 
*need to add redirection when successfully signed up to /login or /


backend/models/user.ts
```diff
@@ -0,0 +1,10 @@
+export interface User {
+    id: string; 
+    email: string; 
+    firstName: string; 
+    lastName: string; 
+    hashedPassword: string; 
+    salt: string; 
+    birthDate: Date; 
+}
+
```
chinkevin25: Did you make this yourself? If so do we need a new interface for this? I thought we would be able to use the Prisma Model for this?

I know Prisma has their own types but i'm not sure how the are generated / extracted: https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#type-definitions

backend/routes/user.ts
```diff
@@ -0,0 +1,28 @@
...
+import app from '../utils/app';
+import prisma from '../utils/db';
+
+const db = prisma; 
+
+app.post('/signup', 
+passport.authenticate('local', {failureRedirect: '/signup', failureMessage: true }),
+ function(req, res) {
+    console.log(req.body)
+ });
```
chinkevin25: I believe you're going to end up moving the route here right, just checking

jessicahe004: i will end up moving this

backend/models/user.ts
jessicahe004: i think i'm just going to delete this file

backend/routes/signUpAuth.ts
```diff
@@ -0,0 +1,47 @@
...
+        if(err) {
+            next(err); 
+        }
+        const user = await db.user.create({
+            data: {
+                email: req.body.email,
+                firstName: req.body.firstName,
+                lastName: req.body.lastName, 
+                password: hashedPassword.toString("base64"),
+                salt: salt //fix salt ??? how to best generate salt using crypto
```
justinpchen94: you already generated it correctly on line 21!

also a cool javascript shorthand is whenever you have an object with fields like this:

```
{
  salt: salt,
}
```

it can be rewritten as

```
{
  salt,
}
```

justinpchen94: see: https://javascript.plainenglish.io/object-literals-using-object-property-shorthand-6360825c60ef#:~:text=When%20ES6%20was%20created%20object,the%20name%20and%20the%20key.&text=myKeyName-,%7D%3B,us%20to%20write%20less%20code!

backend/routes/signUpAuth.ts
```diff
@@ -0,0 +1,47 @@
...
+});
+
+//error caused by emails should be unique but using same email fix error 
+// error states, styling forms, fix routes, deicide what needs to be sent back to the client (what's the next step) (reroute to a specific page), 
+
+// app.post('/signup', 
+// passport.authenticate('local', {failureRedirect: '/signup', failureMessage: true }),
+//  function(req, res) {
+//     console.log(req.body)
+//  });
```
justinpchen94: let's rm if not needed!

backend/utils/app.ts
```diff
@@ -0,0 +1,5 @@
+import express from 'express';
```
justinpchen94: i think this file is not needed since we have a diff app.ts in the backend/ dir already I believe!

backend/utils/db.ts
```diff
@@ -0,0 +1,6 @@
+import { PrismaClient } from '@prisma/client';
```
justinpchen94: same, i think this logic is already in backend/app.ts

jessicahe004: okay ty i will fix that

backend/routes/signUpAuth.ts
```diff
@@ -0,0 +1,36 @@
...
+        const user = await db.user.create({
+            data: {
+                email: req.body.email,
+                firstName: req.body.firstName,
+                lastName: req.body.lastName, 
+                password: hashedPassword.toString("base64"),
+                salt 
+            },
+        })
+        console.log(user); //redirect to the login page
```
justinpchen94, line 33: let's add this as a TODO

backend/routes/signUpAuth.ts
```diff
@@ -0,0 +1,36 @@
...
+                email: req.body.email,
+                firstName: req.body.firstName,
+                lastName: req.body.lastName, 
+                password: hashedPassword.toString("base64"),
+                salt 
+            },
+        })
+        console.log(user); //redirect to the login page 
+    })
+    res.send(200);
```
justinpchen94, line 35: what happens if the user.create fails for some reason?

justinpchen94 (approved):
small nits! nothing blocking on this pr

backend/routes/signUpAuth.ts
jessicahe004: send an error code to frontend and then frontend will issue a warning that says issue with submission try again

chinkevin25 (approved):

PR #15, opened 2023-02-15
T38: Build Sign Up Page

jessicahe004:
Built a functioning form to take in 6 inputs: first name, last name, email, password, and confirm password. 
- Utilized a function handleSubmit to handle the onSubmit event for the form.
- handleSubmit sends an axios post request with the inputs and also handles errors. 
- handleSubmit also makes sure password and confirm password are the same inputs otherwise there is a hidden label state named  setWarning that appears to indicate when that error occurs. 
- added a route on App.tsx for the signup.tsx and updated index.tsx with a browser route. 

*still need to add password validation and refine sending error codes including redirection 
*need to update sign up href link to a route


frontend/src/App.tsx
```diff
@@ -27,11 +29,17 @@ function App() {
...
+</div>
   );
 }
 
 export default App;
+
+{/* <div className="App">
+<h2>{message ?? ""}</h2>
+<button onClick={handleClick}>click me!</button>
+</div> */}
```
chinkevin25: If there is unused code we should strive to remove it

frontend/src/components/signup.css
```diff
@@ -0,0 +1,118 @@
+* {
+    font-family: 'Inter', sans-serif;
+    width: auto; 
+    margin: auto;
+    /* outline: 1px solid red; */
```
chinkevin25: these outlines are for debugging if i remember correctly? Should we remove these as well

frontend/src/components/signup.css
```diff
@@ -0,0 +1,118 @@
...
+
+.weWrapLogo{
+    width: 89px; 
+    height: 103px; 
+    position: absolute;
+    left: 0%;
+    right: 0%;
+    top: 0%;
+    bottom: 0%; 
+}
```
chinkevin25: So one thing we're doing in this CSS page is we're kind of using `classNames` as `ids`

Remember that classNames are meant to be "reusable" across multiple components. As we have it now, it seems like we're just applying a class name as an Id to style it.

>While an ID is specific to a single element, classes can be assigned to multiple elements on a page or throughout the website. They are not unique. And while a single element can only have one ID, it can have multiple classes

We do try to avoid using ids where possible though since css is mostly reusable. 

How could we break this file up to better implement reusable class names

frontend/src/components/signup.css
```diff
@@ -0,0 +1,118 @@
...
+    font-size: 20px; 
+    margin-left: 29em;
+    display: inline-flex; 
+    margin-bottom: 2.25em;
+    flex-direction: row;
+    text-justify: distribute;
+}
+
+
+.password, .firstName, .lastName, .email, .confirmPassword {
```
chinkevin25: so this is a good example of classnames that could be reused. they all share the same class, but they have five different classes! we could consolidate by making a single classname for all of these, maybe something like `.signupFields`

chinkevin25: Alternatively, (not saying this is better or worse) but we could have also taken advantage of other CSS ways 

for example 

`form.signUpForm input { }` would select all inputs in your form with class `signUpForm`

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+        {/* <img className = "weWrapLogo" src = './weWrapGreen.png' alt = 'wewrap logo'> </img> */}
+        <header> 
+        <h3 className = "weWrapTitle"> Sign Up for WeWrap </h3>
+        <p className = "createAccount" > Create a free account or &nbsp;<a href = " ">login </a> </p>
+        </header>
+            {/* <img src = ""> WeWrap Logo </img> */}
+            <form className = "signUpForm" onSubmit = {handleSubmit}> 
+                <label htmlFor = "errorNotification"> {warning} </label>
+                <label htmlFor = "firstName"> 
+                    <p className = "firstNameTitle"> First Name * </p>
```
chinkevin25: I think typically, you don't need a `<p>` tag here. Though I don't know if there are any consequences of having it. so no change needed.

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+        <section className = "container"> 
+        {/* <img className = "weWrapLogo" src = './weWrapGreen.png' alt = 'wewrap logo'> </img> */}
+        <header> 
+        <h3 className = "weWrapTitle"> Sign Up for WeWrap </h3>
+        <p className = "createAccount" > Create a free account or &nbsp;<a href = " ">login </a> </p>
+        </header>
+            {/* <img src = ""> WeWrap Logo </img> */}
+            <form className = "signUpForm" onSubmit = {handleSubmit}> 
+                <label htmlFor = "errorNotification"> {warning} </label>
```
chinkevin25: I don't think we want to be using labels for something that doesnt tie to an input. This seems like it should be a div span or p

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+                <label htmlFor = "firstName"> 
+                    <p className = "firstNameTitle"> First Name * </p>
+                    <input className = "firstName" type = "text" required value = {firstName} onChange = {(event) => setFirstName(event.target.value)}/>
```
chinkevin25: i don't know if TS is smart enough to know this, but if you have an input inside of a label you can usually omit the "for" values.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#:~:text=%3Cinput%3E%20element.-,Alternatively,-%2C%20you%20can%20nest

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+                <p className = "lastNameTitle"> Last Name * </p>
+                    <input className = "lastName" type = "text" required value = {lastName} onChange={(event) => setLastName(event.target.value)}/> 
+                </label>
+                <label htmlFor = "email"> 
+                    <p className = "emailTitle"> Email * </p>
+                    <input className = "email" type = "text" required value = {email} onChange={(event) => setEmail(event.target.value)}/>
+                </label>
+                <label htmlFor = "password"> 
+                    <p className = "passwordTitle"> Password * </p>
+                    <input className = "password" type = "text" required value = {password} onChange={(event) => setPassword(event.target.value)}/>
```
chinkevin25: if we do `type='password'` we get some nice features :)

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    <p className = "firstNameTitle"> First Name * </p>
+                    <input className = "firstName" type = "text" required value = {firstName} onChange = {(event) => setFirstName(event.target.value)}/>
+                </label>
+                <label htmlFor = "lastName">
+                <p className = "lastNameTitle"> Last Name * </p>
+                    <input className = "lastName" type = "text" required value = {lastName} onChange={(event) => setLastName(event.target.value)}/> 
+                </label>
+                <label htmlFor = "email"> 
+                    <p className = "emailTitle"> Email * </p>
+                    <input className = "email" type = "text" required value = {email} onChange={(event) => setEmail(event.target.value)}/>
```
chinkevin25: Do we have plans to do email validation? i.e. making sure that an email is actually being provided rather than just a username like chinkevin25

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+                setWarning('');
+                await axios.post('http://localhost:8000/signup', {
+                    firstName,
+                    lastName,
+                    email,
+                    password, 
+                })
+        
+                .then((res) => {
+                    console.log(res.data)
```
chinkevin25: No change needed for this PR since i think we're going to add more features -- but what should we do after a signup is successful?

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+                })
+        
+                .catch((err) => {
+                    setWarning('Please use a different email.');
+                })
```
chinkevin25: technically, we don't know that a different email is the error -- there could be issues elsewhere. We should be checking the err object to see if there is data that helps us understand what we might want to. I.e. if we have a email collision then maybe this is the right error message.

Maybe there's a bug and somehow a password of length 0 was submitted -- then we might need to surface a different error.

Maybe the login server is down, then we might need to surface something else.

No change needed in this PR but i'd like us to start thinking about what _could_ go wrong.

chinkevin25 (changes requested):

frontend/src/components/signup.tsx
chinkevin25: something else to think about is -- what happens if i tried to maliciously "take" someone elses email. Is that something we should be controlling for? 

What happens if i sign up with jesshe@gmail.com (no clue what your email is)

frontend/src/components/signup.tsx
```diff
@@ -0,0 +1,85 @@
...
+                <label htmlFor = "password"> 
+                    <p className = "passwordTitle"> Password * </p>
+                    <input className = "password" type = "text" required value = {password} onChange={(event) => setPassword(event.target.value)}/> 
+                    <p className = "passwordWarning"> Password must have a minimum of 8 characters and must contain at least one number and one special character. </p>
+                </label>
+                <label htmlFor = "confirmPassword">
+                    <p className = "confirmPasswordTitle"> Confirm Password * </p>
+                    <input className = "confirmPassword" type  = "text" required value = {confirmPassword} onChange={(event) => setConfirmPassword(event.target.value)}/> 
+                </label>
+                <button className = "submitButton" type = "submit"> Sign Up </button>
```
chinkevin25: Forgot to mention: This button right now could be clicked w/ nothing filled out. Should we protect against that?

frontend/src/components/signup.css
```diff
@@ -0,0 +1,118 @@
+* {
+    font-family: 'Inter', sans-serif;
+    width: auto; 
+    margin: auto;
+    /* outline: 1px solid red; */
+}
+
+.container{
+    background-color: #F8FFE5;
+    /* outline: 1px solid red; */
```
justinpchen94: also rm!

frontend/src/components/signup.tsx
jessicahe004: could validation & malicious signing up be resolved with a sent email verification ?

frontend/src/components/signup.tsx
jessicahe004: perhaps send a 200

frontend/src/components/signup.tsx
jessicahe004: the required value prevents that from happening

frontend/src/components/signup.tsx
justinpchen94: well this would be the frontend, so we're not dealing with status codes! is there a sensible _page_ or something that the user should see?

frontend/src/components/signup.tsx
justinpchen94: +1

justinpchen94 (commented):
looks pretty good! just some small things

frontend/src/components/signup.tsx
chinkevin25: Yeah, it's actually kind of an involved process and many ways of doing it. We can leave it as it is for now since I want us to move past Auth soon! But the ideal way of actually doing this would be something like

"Enter Email" hit Submit 
weWrap will them E-Mail the email provided with a link to the rest of signup
User Clicks on a link in the email to continue the rest of the signup process (setting a password)

This way we can be sure that the email is actually the user rather than any one.

Alternatively, we could do whats considered magic link login-- but we dont need to do that for this project. This is where we don't even collect passwords -- we just email them a link and once they click on it, we consider that as signed in.

frontend/src/components/signup.tsx
chinkevin25: yeah TIL! I havent used html forms in a while and I was very surprised that this is the case! Thanks for teaching me something new

frontend/src/components/signup.tsx
chinkevin25: @jesshe004 I think we resolve this and then we're good to go

chinkevin25 (changes requested):

justinpchen94 (approved):
looking good!

jessicahe004:
* still need to password validate

PR #18, opened 2023-02-17
T18: Add a linter

midwestbread:
Start of linter install


backend/package.json
```diff
@@ -22,6 +22,13 @@
     "@types/passport": "^1.0.11",
     "@types/passport-google-oauth20": "^2.0.11",
     "@types/passport-local": "^1.0.35",
+    "@typescript-eslint/eslint-plugin": "^5.52.0",
+    "eslint": "^8.34.0",
+    "eslint-config-standard-with-typescript": "^34.0.0",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-n": "^15.6.1",
+    "eslint-plugin-promise": "^6.1.1",
+    "eslint-plugin-react": "^7.32.2",
```
justinpchen94, line 31: do you know if we need these packages here? looks like we installed these at the root level already?

backend/app.ts
```diff
@@ -1,87 +1,87 @@
...
-                    }
-                })
-                // return user object
-                cb(null, newUser)
-            } else {
-                cb(null, user)
-            }
-        } catch (error) {
-            cb(error, null)
+    if (user == null) { // if user doesn't exist
```
justinpchen94, line 63: hm i think we might want to ensure that the linter is enforcing triple === sign rather than double. not the end of the world if you can't figure it out, just add a TODO and we can investigate later

frontend/package.json
```diff
@@ -25,8 +25,8 @@
   },
   "eslintConfig": {
     "extends": [
-      "react-app",
-      "react-app/jest"
+      "",
+      ""
```
justinpchen94: if this is intentional, let's just remove instead of leaving as empty quotes

package.json
```diff
@@ -0,0 +1,11 @@
+{
```
justinpchen94, line 1: let's try and add a "scripts" section here with a command to run the linter!

justinpchen94 (commented):
looking pretty good! just want a script added and we'll be ready to go

frontend/package.json
midwestbread: This is intentional, if I don't do this, it throws the error:
"ESLint couldn't find the config "react-app" to extend from. Please check that the name of the config is correct.
The config "react-app" was referenced from the config file in "...\togetherwewrap\frontend\package.json"."

I will remove the empty quotes!

backend/package.json
midwestbread: I don't think we need these packages here. These earlier commits were back when I was trying to get the linter running inside the backend folder, but now that we've decided to run the linter in root, all of these changes can be discarded/deleted.

justinpchen94: let's remove then!

PR #20, opened 2023-02-17
T82: Catch the error response for incorrect password entered during login

fahadakhan2:


backend/routes/loginAuth.ts
```diff
@@ -1,46 +1,28 @@
...
-        done(error);
-    }
-}));
-
-export default loginAuthRouter
\ No newline at end of file
+loginAuthRouter.post("/password", function (req, res, next) {
+    passport.authenticate("local", function (err, user, info) {
+      if (err) {
+          // Replace this with Error Message
```
justinpchen94: comment still needed?

backend/routes/loginAuth.ts
```diff
@@ -1,46 +1,28 @@
...
+        return res.status(401).send('Email or password did not match. Please try again.')
+      }
+      if (!user) {
+          // Replace this with Error Message
```
justinpchen94: same

backend/routes/loginAuth.ts
```diff
@@ -1,46 +1,28 @@
...
+        return res.status(401).send('Email or password did not match. Please try again.');
+      }
+      // NEED TO CALL req.login()!!!
```
justinpchen94: ah was this what the bug was?

frontend/src/Components/loginForm.tsx
```diff
@@ -2,21 +2,25 @@ import React, { useState } from 'react';
...
     const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
         event.preventDefault();
         try {
           const response = await axios.post("http://localhost:8000/login/password", {
             email,
             password,
           });
-          console.log(response.data)
           // TODO T105 (FK) After successful response redirect user to whatever page they're trying to reach
+          console.log(response.data)
```
justinpchen94: is this a debug console log? let's rm if it isn't necessary

frontend/src/Components/loginForm.tsx
```diff
@@ -35,6 +39,7 @@ export const LoginForm = () => {
           {/* T80 (FK): Add route to sign up page once sign up code is push and merged */}
           <p className='subheader'>Don't have an account?<a href='' > Sign up</a> </p>
           <form action="/login/password" method="post" onSubmit={handleSubmit}>
+              <p className='error_message'>{errorMessage}</p>
```
justinpchen94: a common tactic here is to conditionally render the whole element, rather than leaving this as is

```
{errorMessage && <p className='error_message'>{errorMessage}</p>}
```

reason being that even if errorMessage is an empty string, the `p` element might still have styling that affects the rest of the page

backend/routes/loginAuth.ts
chinkevin25: TLDR passport.authenticate does somethign weird (or the way we set it up is weird) and it was returning HTML with the error message. 

we need to call req.login() because we're no longer using the default callback.

chinkevin25 (commented):
Looks good pending Justins comments

frontend/src/Components/loginForm.tsx
fahadakhan2: yes will remove.

frontend/src/Components/loginForm.tsx
fahadakhan2: Okay I see. Changed!

justinpchen94 (approved):

PR #21, opened 2023-02-17
T59: Add FB OAuth strategy from passport

kdata12:
Added [Passport](http://passportjs.org/) strategy for authenticating with [Facebook](http://www.facebook.com/) using the OAuth 2.0 API.

http://www.passportjs.org/packages/passport-facebook/

- Added /auth/facebook route with passport facebook middleware
- Moved ENV variables to /utils/config.ts
- Move passport local strategy to app.ts
- Added facebookID column into prisma schema


backend/app.ts
```diff
@@ -42,12 +48,13 @@ passport.deserializeUser(async (id: string, done: any) => {
         }
     })
     console.log('found user from cookie')
+    console.log(user)
```
chinkevin25: nit: are these debug logs?

backend/app.ts
```diff
@@ -80,8 +87,67 @@ passport.use(new GoogleStrategy({
...
+        const hashedPassword = crypto.pbkdf2Sync(password, user.salt, 310000, 32, 'sha256').toString('hex');
+        if (user.password !== hashedPassword) {
+            return done(null, false, { message: 'Incorrect email or password.' });
+        }
+
+        return done(null, user);
+
+    } catch (error) {
+        done(error);
+    }
```
chinkevin25: Since you're FLYING and implementing a shit ton of OAUTH. Challenge for you! We're repeating a lot of Code between Google OAUTH, FB OAUTH and local strategy. How could we DRY the code up?

backend/routes/loginAuth.ts
```diff
@@ -1,11 +1,10 @@
...
 const loginAuthRouter = express.Router()
-import { Strategy as LocalStrategy } from 'passport-local';
 import passport from 'passport'
-import prisma from '../utils/prismaClient'
-import crypto from 'crypto';
 
-const db = prisma;
+loginAuthRouter.get('/', (req, res) => {
+    res.status(400).send('failed login')
+})
```
chinkevin25: Whats this for?

backend/app.ts
kdata12: Yup, I just removed them.

backend/routes/loginAuth.ts
kdata12: Oh this was just to test local strategy `failureRedirect: '/login'`, I will remove it.

backend/app.ts
kdata12: One method that I can think of is to find the implementation similarities between those passport auth strategies and then abstract the user finding & creation logic to some helper functions, and reuse those functions inside our passport strategies.

kdata12: For Google and FB OAUTH, I find that they very similar implementations and I can try to refactor them. As for the local strategy, I'm only seeing 1 implementation to authenticate user login with email and hashedpassword, unless we need to reuse this local strategy logic in the future. Can you give me pointers?

chinkevin25 (approved):

PR #22, opened 2023-02-20
T46: Implement Sessions

kdata12:
Implemented sessions middleware using prisma session store: https://github.com/kleydon/prisma-session-store

I also created a session schema to store user's sessions.

Cookie's max age is set to 3 days and with each session touch, the cookie max age will reset back to 3 days.


backend/app.ts
```diff
@@ -2,34 +2,55 @@ import * as dotenv from 'dotenv';
...
 app.use(cors());
 app.use(express.json());
 
-const secretcode = process.env.SESSION_SECRET as string;
-const clientID = process.env.GOOGLE_CLIENT_ID as string;
-const clientSecret = process.env.GOOGLE_CLIENT_SECRET as string;
+app.use(
+    expressSession({
+        cookie: {
+            maxAge:  20 * 1000
```
justinpchen94: it's a good practice to do either of the following:

1) store a descriptive const variable with this number at the very top and use it below
e.g.
```
const HOURS_IN_MS = 60 * 60 * 1000
```

or

2) leave the duration in a comment
e.g.
```
maxAge: 20 * 1000 // 20 seconds
```

backend/app.ts
```diff
@@ -2,34 +2,55 @@ import * as dotenv from 'dotenv';
...
+        },
+        secret: secretcode,
+        resave: false,
+        saveUninitialized: false,
+        store: new PrismaSessionStore(
+            new PrismaClient(),
+            {
+                checkPeriod: 2 * 60 * 1000,
```
justinpchen94: same here

backend/app.ts
```diff
@@ -47,19 +68,20 @@ passport.deserializeUser(async (id: string, done: any) => {
         }
     })
     console.log('found user from cookie')
+    console.log(user)
```
justinpchen94: debug comment?

backend/app.ts
```diff
@@ -85,11 +107,49 @@ passport.use(new GoogleStrategy({
...
 
-    passport.use(new LocalStrategy({
-                usernameField: 'email',
-                passwordField: 'password'
-        },
-        async (email: string, password: string, done: Function) => {
+passport.use(new FacebookStrategy({
+    clientID: facebookClientID,
+    clientSecret: facebookAppSecret,
+    callbackURL: "http://localhost:8000/auth/facebook/callback",
```
justinpchen94: no change needed but just keep in mind this URL will have to eventually be abstracted out, since we won't be using locahost after we deploy this code!

backend/app.ts
```diff
@@ -85,11 +107,49 @@ passport.use(new GoogleStrategy({
...
+        } catch (error) {
+            cb(error, null)
+        }
+    }
+));
+
+app.use('/', testRouter)
+app.use('/auth/google', googleOAuthRouter)
+app.use('/auth/facebook', facebookOAuthRouter)
+app.use('/login', loginAuth)
```
justinpchen94: this looks copy pasted from above? rm?

backend/app.ts
justinpchen94: how do you think we would solve this moving forward, so we can stop hardcoding in the localhost URL?

justinpchen94 (commented):
couple comments!

backend/app.ts
kdata12: removed!

backend/app.ts
```diff
@@ -2,38 +2,60 @@ import * as dotenv from 'dotenv';
...
+        },
+        secret: secretcode,
+        resave: false,
+        saveUninitialized: false,
+        store: new PrismaSessionStore(
+            new PrismaClient(),
+            {
+                checkPeriod: 2 * 60 * 1000, //2 minutes
+                dbRecordIdIsSessionId: true,
+                dbRecordIdFunction: undefined,
```
justinpchen94, line 47: should this be undefined? if this is undefined, do we need to keep it here?

backend/app.ts
kdata12: yes!

backend/app.ts
kdata12: My guess is to use an ENV variable?

justinpchen94: correct! you'd pass in the API_URL or similar as an env var

backend/app.ts
kdata12: dbRecordIdFunction: A function to generate the Prisma Record ID for a given session ID.

I just kept it there just to be explicit.

backend/app.ts
```diff
@@ -2,40 +2,59 @@ import * as dotenv from 'dotenv';
...
+import { PrismaClient } from '@prisma/client';
+import facebookStrategy from 'passport-facebook';
+import facebookOAuthRouter from './routes/facebookOAuth';
 import { Strategy as LocalStrategy } from 'passport-local';
 import crypto from 'crypto';
+import { secretcode, googleClientID, googleClientSecret, facebookAppSecret, facebookClientID } from './utils/config'
+
 dotenv.config();
 const GoogleStrategy = googleStrategy.Strategy;
+const FacebookStrategy = facebookStrategy.Strategy;
```
chinkevin25, line 22: Seems like we have the FB OAuth changes on this PR as well. Do you want to try to separate these PRs?

backend/app.ts
```diff
@@ -2,40 +2,59 @@ import * as dotenv from 'dotenv';
...
-
-app.use(session({
-    secret: secretcode,
-    resave: true,
-    saveUninitialized: false,
-}));
+app.use(
+    expressSession({
+        cookie: {
+            maxAge: 3 * 24 * 60 * 60 * 1000 //3 days
```
chinkevin25: Let's make this a variable rather than a magic number w/ comments

backend/app.ts
```diff
@@ -2,40 +2,59 @@ import * as dotenv from 'dotenv';
...
+        },
+        secret: secretcode,
+        resave: false,
+        saveUninitialized: false,
+        store: new PrismaSessionStore(
+            new PrismaClient(),
+            {
+                checkPeriod: 2 * 60 * 1000, //2 minutes
```
chinkevin25: Let's make this a variable rather than a magic number w/ comments

chinkevin25 (changes requested):

backend/app.ts
kdata12: Ok I will remove FB OAuth from this PR

kdata12: Earlier I squashed and merged my FB OAuth branch into main with your approval, then a moment later I pulled main in this branch which has the FB OAuth codes.

backend/.sample-env
```diff
@@ -3,4 +3,9 @@ GOOGLE_CLIENT_ID=""
...
 SESSION_SECRET=""
 FACEBOOK_CLIENT_ID=""
-FACEBOOK_APP_SECRET=""
\ No newline at end of file
+<<<<<<< HEAD
+FACEBOOK_APP_SECRET=""
+FACEBOOK_CALLBACK_URL=""
+=======
+FACEBOOK_APP_SECRET=""
+>>>>>>> 381b6f9e10000e9a6a4785330d211bc08281bc8b
```
chinkevin25: Looks like we have a merge conflict that didn't get resolved!

backend/app.ts
```diff
@@ -2,44 +2,59 @@ import * as dotenv from 'dotenv';
...
+import crypto from 'crypto';
+import { secretcode, googleClientID, googleClientSecret, facebookAppSecret, facebookClientID, facebookCallBackURL, googleCallBackURL} from './utils/config'
 import facebookStrategy from 'passport-facebook';
 import facebookOAuthRouter from './routes/facebookOAuth';
 import loginAuth from './routes/loginAuth'
-import { Strategy as LocalStrategy } from 'passport-local';
-import crypto from 'crypto';
-import { secretcode, googleClientID, googleClientSecret, facebookAppSecret, facebookClientID } from './utils/config'
-
+dotenv.config();
```
chinkevin25: I vaguely remember that we need dotenv.config() to be ran as like one of the first few lines. Should this be higher up? (might not have problems now)

backend/app.ts
chinkevin25: I think Justin meant that this should be removed

backend/prisma/schema.prisma
```diff
@@ -18,3 +18,10 @@ model User {
   salt        String?
   facebookID String?   @unique
 }
+
+model Session {
+  id        String   @id
+  sid       String   @unique
```
chinkevin25, line 24: question for you, do we need to maintain both id and sid? in theory, these could be the same value right? No change needed though cause I think this is fairly common

chinkevin25 (changes requested):

backend/prisma/schema.prisma
kdata12: This is a good point. So I actually tested this out by removing the id column and I got this error from the prisma client:

image

So id and sid are indeed required in our session table!

backend/app.ts
kdata12: Ok fixed! I move it up higher just incase.

chinkevin25 (approved):

PR #26, opened 2023-02-21
T64: Add backend logic for logout

fahadakhan2:
When the /logout route is called, it will first call req.logout() which will remove the user from the req object's session data. It will then call req.session.destroy() which will destroy the session data and remove the session from the DB.


backend/routes/logout.ts
```diff
@@ -0,0 +1,20 @@
...
+logoutRouter.post('/logout', function(req, res, next){
+    req.logout(function(err) {
+      if (err) {
+        return next(err);
+      }
+      req.session.destroy(function(err) {
+        if (err) {
+          return next(err);
+        }
+        res.redirect('/login');
```
justinpchen94: nit: let's make this go to home page!

justinpchen94 (approved):
just update the redirect and we're good!

backend/routes/logout.ts
fahadakhan2: home page meaning the tempLandingPage I added?

fahadakhan2:
When the /logout route is called, it will first call req.logout() which will remove the user from the req object's session data. It will then call req.session.destroy() which will destroy the session data and remove the session from the DB.

PR #28, opened 2023-02-23
T103: Add Google and Facebook Login Button in Frontend

kdata12:
### - Added API call to backend to take user through OAuth process.

image


frontend/src/components/loginForm.tsx
```diff
@@ -58,6 +68,17 @@ export const LoginForm = () => {
...
+                <p>OR</p>
+              </div>
+              <div className='googleContainer' onClick={handleGoogleOAuth}>
+                <img src={googleIcon} alt='googleIcon'/>
+                <p>Log in with Google</p>
+              </div>
+              <div className='facebookContainer' onClick={handleFaceBookOAuth}>
+                <img src={facebookIcon} alt='googleIcon'/>
+                <p>Log in with Facebook</p>
+              </div>
```
chinkevin25: We're kind of using a div as a button. While that functionally works -- it's not a great practice.

If this is going to navigate us to a different URL we should probably be using links (`a`) but styled as a button

frontend/src/components/loginForm.tsx
```diff
@@ -33,6 +35,14 @@ export const LoginForm = () => {
     setPassword(event.target.value);
   };
 
+  const handleGoogleOAuth = () => {
+    window.open('http://localhost:8000/auth/google', '_self')
+  }
+
+  const handleFaceBookOAuth = () => {
+    window.open('http://localhost:8000/auth/facebook', '_self')
+  }
```
chinkevin25: This works -- but if we change our div to an `a` tag and then use the href and target field.

chinkevin25: Do we even need to navigate to a different page though? Doesn't the google/fb OAuth flow take you to like a "sign in via google / fb" page that you can then supply a success and fail redirect path?

chinkevin25 (changes requested):
I have some questions about the OAuth flow (caveat I didn't try actually signing in via OAuth so I don't exactly how it's working).

frontend/src/components/loginForm.tsx
kdata12: fixed!

frontend/src/components/loginForm.tsx
kdata12: Yeah so I tried sending an GET request to the our backend google and facebook endpoint but it doesn't work and it gave me a CORS error. I think this is because OAuth doesn't support CORS ?  (https://stackoverflow.com/questions/55079181/what-is-the-best-practice-to-use-oauth2-react-node-js-and-passport-js-to-authe/63736799#63736799)

The only solution I've found so far is to do a redirect to the our OAuth endpoint directly.

chinkevin25: https://create-react-app.dev/docs/proxying-api-requests-in-development/

I was doing some more research and i think this is why we're not able to proxy our requests properly! Do you want to see if this solves our problem?

chinkevin25 (approved):

frontend/src/components/loginForm.tsx
kdata12: I noticed that I didn't send you the CORS error message. Here's an axios GET request to `localhost:8000/auth/google `(backend) and it gave me this error. This is for reference.

image

kdata12: Also I tried proxying our requests to our backend by putting `"proxy": "http://localhost:8000"` in package.json as stated from the react doc you linked. Then I tried doing an axios GET request to`'http://localhost:8000/auth/google'`. I'm not sure how proxy works tbh.

image

chinkevin25: I think if we have the proxy working, we don't need to specify `http://localhost:8000` anymore, just `/auth/google`.

The proxy works by sending any requests that originate in `localhost:3000` to `localhost:8000` (proxy server). If we're still getting CORS errors, then we might want to enable CORS in our headers if we need to.

kdata12: So I got the same CORS error message when I use `axios.get('/auth/google')` with `localhost:8000` as the proxy:

image

It seems like its trying to request to `localhost:3000/auth/google` instead of `localhost:8000/auth/google` with the proxy turn on. 

Network:
image

PR #29, opened 2023-02-24
T18: Add a linter (2nd attempt)

midwestbread:
this is up and running again, I didn't manually lint this time because I'm 4 commits behind main so I think manually linting would just cause a bunch of merge conflicts, hopefully we can merge and then do it this time.


tsconfig.json
```diff
@@ -0,0 +1,99 @@
...
+        // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
+        // "typeRoots": [],                                  /* Specify multiple folders that act like './node_modules/@types'. */
+        // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
+        // "allowUmdGlobalAccess": true,                     /* Allow accessing UMD globals from modules. */
+        // "moduleSuffixes": [],                             /* List of file name suffixes to search when resolving a module. */
+        // "resolveJsonModule": true,                        /* Enable importing .json files. */
+        // "noResolve": true,                                /* Disallow 'import's, 'require's or '<reference>'s from expanding the number of files TypeScript should add to a project. */
+        /* JavaScript Support */
+        "allowJs": true, /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */
```
justinpchen94: nit: let's set this to false! we should only be using ts files ideally

justinpchen94 (commented):
i think this looks good! but gonna have @chinkevin25 take a pass too

justinpchen94:
Let's also update the title to use the task title so the format of T#: Task Title

package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^5.53.0",
+    "eslint": "^8.34.0",
+    "eslint-config-standard-with-typescript": "^34.0.0",
+    "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-n": "^15.6.1",
+    "eslint-plugin-promise": "^6.1.1",
+    "eslint-plugin-react": "^7.32.2",
```
chinkevin25, line 9: I'm not 100% sure if these are being used? I would imagine they have to be imported / configured in your `.eslintrc.js` but it's possible that you're using  something that already includes these and these are just peer dependencies.

package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "devDependencies": {
+    "@typescript-eslint/eslint-plugin": "^5.53.0",
```
chinkevin25, line 3: do we need to specify the usage of this anywhere? The docs seem to mention something about setting this as the parser? do you know if that is required?

chinkevin25 (commented):
Looks great just want to make sure we're using these rules!

chinkevin25 (approved):
Approving since it works as is

PR #32, opened 2023-03-03
T70: Extend User Table to store profile photo url & T73: Store venmo/cash/paypal profile urls on user table

jessicahe004:
Added columns on the Users Table: venmo/cashapp/paypal. 
Created a profiles model using a one-to-one relation.


chinkevin25 (approved):
Looks good -- minor nit of trying to keep the 1PR per Task habit up if we can! These are low line number tasks so I get why you combined them

PR #33, opened 2023-03-03
T86: Handle Registration Errors & T85: SignUp Password Validation 

jessicahe004:
*in progress* 

[T85] pending review - Added a minimum length checker and a number for password validation and a special character.


frontend/src/components/signup.tsx
```diff
@@ -12,13 +12,23 @@ export const SignUp = () => {
     const [lastName, setLastName] = useState<string>('');
     const [email, setEmail] = useState<string>('');
     const [password, setPassword] = useState<string>('');
+    const [validLength, setValidLength] = useState<boolean>(false); 
+    const [hasNumber, setHasNumber] = useState<boolean>(false); 
+    const [specialChar, setSpecialChar] = useState<boolean>(false); 
     const [confirmPassword, setConfirmPassword] = useState<string>('');
     const [warning, setWarning] = useState<string>(''); 
+    const regex = /[\!\@\#\$\%\^\&\*\)\(\+\=\.\<\>\{\}\[\]\:\;\'\"\|\~\`\_\-]/g;
```
justinpchen94: let's use a more descriptive name for this var! reading this alone would tell me basically nothing about what regex pattern this is!

frontend/src/components/signup.tsx
```diff
@@ -12,13 +12,23 @@ export const SignUp = () => {
...
 
     const handleSubmit = async (e: React.FormEvent<HTMLFormElement>) => {
 
         e.preventDefault(); 
 
+        setValidLength(password.length >= 8 ? true : false);
```
justinpchen94: nit: this can simply be `setValidLength(password.length >= 8)`

frontend/src/components/signup.tsx
```diff
@@ -12,13 +12,23 @@ export const SignUp = () => {
...
+        setSpecialChar(regex.test(password));
+        setHasNumber(/\D/.test(password));
+
+        if(validLength && hasNumber && specialChar) {
```
justinpchen94: i am not exactly sure of what will happen here but there is a possibility for a race condition here. these state variables are relying on the calls you have on line 26-28 above to correctly finish setting state _before_ this line executes. you can extract these variables to rely on the current `password` rather than reference the state vars instead

```
if (password.length >= 8 && regex.test(password) && /\D/.test(password)) {
```

but with the vars extracted out cleanly!

frontend/src/components/signup.tsx
```diff
@@ -27,22 +37,28 @@ export const SignUp = () => {
                     email,
                     password, 
                 })
-        
+
```
justinpchen94: nit: generally with chained promises you do want to have them together and not separate with the new line

frontend/src/components/signup.tsx
```diff
@@ -27,22 +37,28 @@ export const SignUp = () => {
...
                 .then((res) => {
                     setWarning('Successful submission.');
                 })
-        
+
```
justinpchen94: same as above

frontend/src/components/signup.tsx
```diff
@@ -12,13 +12,23 @@ export const SignUp = () => {
     const [lastName, setLastName] = useState<string>('');
     const [email, setEmail] = useState<string>('');
     const [password, setPassword] = useState<string>('');
+    const [validLength, setValidLength] = useState<boolean>(false); 
+    const [hasNumber, setHasNumber] = useState<boolean>(false); 
+    const [specialChar, setSpecialChar] = useState<boolean>(false);
```
justinpchen94: generally good rule of thumb for variables that are meant to hold boolean vals is to make the var name as if it's based on the question.

`isvalidLength` vs `validLength`
`hasNumber` is good!
`hasSpecialChar` vs `specialChar`

something like `specialChar` is a little ambiguous. is it a string var holding a special char? does it transform special chars in a string? etc.

justinpchen94 (commented):
couple comments!

kdata12:
After you resolve the merge conflict, I can go ahead and approve this PR

kdata12 (commented):
I think your CSS is fine here but in the future, we should use module.css to prevent css from other component from interfering with each other. You don't have to change it in this PR.

https://www.gatsbyjs.com/docs/how-to/styling/css-modules/

kdata12:
LGTM

kdata12 (approved):

PR #34, opened 2023-03-08
T107: contact form backend

fahadakhan2:
## Title: T107 contact form backend
Screenshot 2023-04-01 at 10 36 44 PM

## Description 
Upon creating a contact by submitting the form the contacts info will be shown in the UI for /contactPage.

Backend logic for contact creation by the user. Reads the form and stores contact information in the database correctly. For relationships and importantEventDates for now it is a string that is inputted by the user. Did not add dropdowns or calender select for the dates. Can add this later on.

## Types of changes

## Checklist

- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules


backend/routes/contactCreator.ts
```diff
@@ -0,0 +1,45 @@
...
+            })
+        }
+        for (const importantDate of req.body.importantDates) {
+            await db.importantDateEvent.create({
+                data: {
+                    date: importantDate.date,
+                    eventType: importantDate.event,
+                    contact: { connect: { id: contact.id } }
+                }
+            })
```
justinpchen94: nit: we should be using `createMany`  instead of looping https://www.prisma.io/docs/concepts/components/prisma-client/crud#create-multiple-records

justinpchen94 (approved):

frontend/src/components/contactsList.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    const handleContactCreate = (newContact: Contact) => {
+        setContacts((prevState) => [...prevState, newContact])
+    }
+
+    useEffect(() => {
+        const getContacts = async () => {
+            try {
+                const response = await axios.get('http://localhost:8000/contacts')
```
kdata12: Was this routed created? If it is, is it returning contact data correctly?

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+  const [errorMessage, setErrorMessage] = useState<string>('')
+  const [error, setError] = useState<boolean>(false)
+  const [maxFriends] = useState<number>(3)
+  const [eventType, setEventType] = useState<EventType>()
+
+  const handleSpecialPersonChange = (friend: Friend): void => {
+    setSpecialPerson(friend)
+  }
+
+  const handleSpecialPersonRemove = (friend: Friend[] | undefined): void => {
```
chinkevin25: as a type -- you could do `friend?: Friend[]` instead of doing `| undefined`

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+    friend !== undefined && setSpecialPerson(friend[0])
```
chinkevin25: is this type safe? we check to make sure that the friend input is defined, but we technically don't know that `friend[0]` exists is this a problem?

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+    setStartDate(event.target.value)
+  }
+
+  const handleEndDateChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    setEndDate(event.target.value)
+  }
+
+  const handleSpecialDateChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    setSpecialDate(event.target.value)
+  }
```
chinkevin25: nit, no change needed but we do very similar logic here.

We could dry this up and make it all one function that is like just `handleChange`

but some people argue that we shouldn't do that because it's harder to modify this code if we need to process the event.target.value first.

No change needed just wanted us to think about it a little!

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,237 @@
...
+    setEventType(e.target.value as EventType)
+  }
+
+  const handleFriendsChange = (friend: Friend): void => {
+    if (friends.length >= maxFriends) {
+      handleError(`Only a max of ${maxFriends} friends are allowed`)
+    } else if (!friends.includes(friend)) {
+      setFriends(friends => [...friends, friend])
+    }
+  }
```
chinkevin25: Here's a fun thought. No change needed either but This works but it's "inefficient". It doesn't matter cause the length of our friends array is not large enough to matter, but this is a O(N) way of doing it.

How else could we implement this?

One way is to use a Set! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

frontend/src/components/SearchBarFilter.tsx
chinkevin25, line 1: nice! this component looks very useful! but it also looks like it should be generic! Right now there's a lot of language specific to friends but seems like this is just a good way of showing an array of data and filtering! How could we make this more generic so another developer wants to use this?

backend/prisma/schema.prisma
```diff
@@ -102,6 +106,53 @@ enum InviteStatus {
...
+  Plan               Plan?                     @relation(fields: [planId], references: [id])
+  planId             String?
+}
+
+model ImportantDateEvent {
+  id        String  @id @default(uuid())
+  contact   Contact @relation(fields: [contactID], references: [id])
+  contactID String
+  date      String
+  eventType String
```
chinkevin25, line 106: this should be an ENUM right?  You already have it too i think

frontend/src/components/contactsList.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  const handleContactCreate = (newContact: Contact) => {
+    console.log(newContact)
+    setContacts((prevState) => [...prevState, newContact])
+  }
+
+  useEffect(() => {
+    const getContacts = async () => {
+      try {
+        const response = await axios.get('http://localhost:8000/contacts')
```
chinkevin25: we're missing this route still  I think. If you want to do it in a separate PR can you slap a TODO comment above this line at the very least.

chinkevin25 (changes requested):

.eslintrc.js
```diff
@@ -14,9 +14,14 @@ module.exports = {
     project: true,
     tsconfigRootDir: __dirname
   },
-
+  plugins: [
+  ],
```
justinpchen94: is this required to be here if it's empty?

.eslintrc.js
```diff
@@ -14,9 +14,14 @@ module.exports = {
...
   rules: {
     eqeqeq: ['error', 'always'],
+    '@typescript-eslint/consistent-type-definitions': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/indent': 'off',
```
justinpchen94: honestly this one is actually a helpful one to keep - any reason you think we should not enforce indentation depths?

backend/app.ts
```diff
@@ -65,18 +69,19 @@ passport.serializeUser((user: any, done: any) => {
 passport.deserializeUser(async (id: string, done: any) => {
     const user = await db.user.findFirst({
         where: {
-            id: id
+            id
         }
     })
-    return done(null, user)
+    console.error(user)
```
justinpchen94: any reason this is an error log?

backend/app.ts
```diff
@@ -145,24 +150,21 @@ passport.use(
...
-                    email: email
-                },
+                    email
+                }
             })
 
-            if (!user || !user.salt) {
-                return done(null, false);
+            if (!user?.salt) {
+                done(null, false); return;
```
justinpchen94: this looks like a typo?

backend/package.json
```diff
@@ -51,4 +51,4 @@
   "prisma": {
     "seed": "ts-node ./utils/userContactPlanSeed.ts"
   }
-}
+}
```
justinpchen94: let's keep the new line at the end of files! this is a standard convention

backend/prisma/schema.prisma
```diff
@@ -70,6 +71,8 @@ model Plan {
   createdAt        DateTime         @default(now())
   updatedAt        DateTime         @updatedAt
   PlanMembership   PlanMembership[]
+  Contact          Contact[]
```
justinpchen94: is this field the same as `contacts`?

backend/prisma/schema.prisma
```diff
@@ -102,6 +106,53 @@ enum InviteStatus {
...
+  firstName          String
+  lastName           String?
+  phoneNumber        String?
+  notes              String?
+  source             String?
+  ownerID            String
+  owner              User                      @relation(fields: [ownerID], references: [id])
+  importantDateEvent ImportantDateEvent[]
+  relationships      UserContactRelationship[]
+  Plan               Plan?                     @relation(fields: [planId], references: [id])
```
justinpchen94: the fact taht this field is of type `Plan?` instead of `Plan[]?` implies to me that a contact can only be a participant in a max of one Plan. is this the constraint we want to apply?

backend/prisma/schema.prisma
```diff
@@ -102,6 +106,53 @@ enum InviteStatus {
...
+  contactID String
+  date      String
+  eventType String
+}
+
+model UserContactRelationship {
+  id               String  @id @default(uuid())
+  contact          Contact @relation(fields: [contactID], references: [id])
+  contactID        String
+  relationshipType String
```
justinpchen94, line 113: should this be an enum?

backend/prisma/schema.prisma
```diff
@@ -102,6 +106,53 @@ enum InviteStatus {
...
+}
+
+model GiftChoice {
+  id          String   @id @default(uuid())
+  Pledge      Pledge   @relation(fields: [pledgeID], references: [id])
+  pledgeID    String
+  giftUrl     String
+  giftName    String
+  dateCreated DateTime @default(now())
```
justinpchen94: nit: general naming convention is `createdAt`

backend/routes/contactCreator.ts
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
+import { type User } from '@prisma/client'
+import express from 'express'
+import prisma from '../utils/prismaClient'
+
+const contactCreatorRouter = express.Router()
+const db = prisma
+
+contactCreatorRouter.post('/', async function (req, res) {
+    console.log(req.body)
```
justinpchen94: should we keep this?

backend/routes/loginAuth.ts
```diff
@@ -1,26 +1,21 @@
-import { type User } from "@prisma/client";
-import express from "express";
+import { type User } from '@prisma/client';
+import express from 'express';
 import passport from 'passport'
 import prisma from '../utils/prismaClient'
 
 const loginAuthRouter = express.Router()
-const db = prisma;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
justinpchen94: should we disable this line or should we actually think about why we want to keep an unused variable?

backend/routes/loginAuth.ts
```diff
@@ -1,26 +1,21 @@
...
+    res.render('login')
+  })
 
-loginAuthRouter.post("/password", passport.authenticate('local'),
+loginAuthRouter.post('/password', passport.authenticate('local'),
   function (req, res) {
-    console.info('user from local auth: ' + (req.user as User).id);
-    res.redirect('/'); //this is not really working, the redirect is still coming from the client
-  });
+    console.log('user from local auth: ' + (req.user as User).id)
```
justinpchen94: generally we use `info` logs for informational logs that we actually want to keep around - is this console log something you're using for debugging or should we keep this log in there?

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+  const [relationships, setRelationships] = useState<Relationship[]>([])
+  const [importantDates, setImportantDates] = useState<ImportantDate[]>([])
+  const [eventDate, setEventDate] = useState<string>('')
+  const [eventType, setEventType] = useState<string>('')
+  const [relationshipType, setRelationshipType] = useState<string>('')
+  const [email, setEmail] = useState<string>('')
+  const [phoneNumber, setPhoneNumber] = useState<string>('')
+  const [notes, setNotes] = useState<string>('')
+  const [errorMessage, setErrorMessage] = useState<string>('')
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
justinpchen94: same here - should we be commenting out this line if we're declaring variables that aren't being used?

frontend/src/components/contactsList.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const ContactsList = () => {
+  const [showCreateAContactForm, setShowCreateAContactForm] = useState<boolean>(false)
+  const [contacts, setContacts] = useState<Contact[]>([])
+
+  const toggleContactForm = () => {
+    setShowCreateAContactForm(!showCreateAContactForm)
+  }
+
+  const handleContactCreate = (newContact: Contact) => {
+    console.log(newContact)
```
justinpchen94: rm or turn it into a console.info?

frontend/src/components/contactsList.tsx
```diff
@@ -0,0 +1,90 @@
...
+    const getContacts = async () => {
+      try {
+        const response = await axios.get('http://localhost:8000/contacts')
+        const contactsData = response.data.contacts as Contact[]
+        setContacts(contactsData)
+      } catch (error) {
+        console.error(error)
+      }
+    }
+    void getContacts()
```
justinpchen94: is this intentional? i've never seen this syntax before haha

frontend/src/components/contactsList.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  useEffect(() => {
+    const getContacts = async () => {
+      try {
+        const response = await axios.get('http://localhost:8000/contacts')
+        const contactsData = response.data.contacts as Contact[]
+        setContacts(contactsData)
+      } catch (error) {
+        console.error(error)
```
justinpchen94, line 47: should we be doing something else besides logging the error to console?

frontend/src/components/contactsList.tsx
fahadakhan2: Yeah it is.

frontend/src/components/contactsList.tsx
fahadakhan2: I'll do it in this PR.

.eslintrc.js
fahadakhan2: removed.

.eslintrc.js
fahadakhan2: Will keep

backend/app.ts
fahadakhan2: removed it

backend/app.ts
fahadakhan2: When I leave it like this:
if (!user.salt) {
      done(null, false);

I get this error for the user word -> 'user' is possibly 'null'.

fahadakhan2: Would including a check if the user exists at all be a good fix such as:

if (!user) {
      done(null, false);
      return;
}
    
But then this gives me an error below in the argument 'user.salt' passed to hashedPassword.
error = Argument of type 'string | null' is not assignable to parameter of type 'BinaryLike'.
  Type 'null' is not assignable to type 'BinaryLike'.
  
 const hashedPassword = crypto.pbkdf2Sync(password, **user.salt**, 310000, 32, 'sha256').toString('base64')

backend/package.json
fahadakhan2: Im not sure why but when I add the extra space and try to save the file its reverting back and deleting the extra line that I add.

backend/prisma/schema.prisma
fahadakhan2: not sure what you mean?

backend/prisma/schema.prisma
fahadakhan2: I'm not sure originally I had it as an enum but since I was getting confused on its implementation I just used String type field.

backend/routes/contactCreator.ts
fahadakhan2: nope removed it!

backend/routes/loginAuth.ts
fahadakhan2: I ended up deleting it and the prisma import since it doesn't seem necessary on this page.

backend/routes/loginAuth.ts
fahadakhan2: yeah it was being used for debugging I'll go ahead and remove.

frontend/src/components/contactsForm.tsx
fahadakhan2: I am using the setErrorMessage so I put the line below in my return statement. So hopefully that's valid.
Screenshot 2023-04-13 at 12 45 53 PM

frontend/src/components/contactsList.tsx
fahadakhan2: will remove!

frontend/src/components/contactsList.tsx
fahadakhan2: I did this based off the error message when I have getContacts() written on line 52.
Error:
Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.

frontend/src/components/contactsList.tsx
fahadakhan2: Probably display an error message to the user!

frontend/src/components/contactsList.tsx
fahadakhan2: I sent you the code on discord

frontend/src/components/PlanForm.tsx
fahadakhan2: this is for hyun he's working on this

backend/prisma/schema.prisma
fahadakhan2: Right now I only have it as a string that is inputted by the user. I was confused on how to implement it using ENUM.

backend/prisma/schema.prisma
fahadakhan2: No it is not but I think Hyun might've already changed this in his code.

backend/app.ts
fahadakhan2: Update: This was fixed after I pulled hyuns latest push to main.

backend/prisma/schema.prisma
fahadakhan2: should I change the capital Contact in the Plan model to lowercase to match the users model? We should prob establish how we want to capitalize the "label" in our db models.

kdata12: The convention is camel case, so lets use `contacts` here

backend/prisma/schema.prisma
```diff
@@ -70,6 +71,8 @@ model Plan {
   createdAt        DateTime         @default(now())
   updatedAt        DateTime         @updatedAt
   PlanMembership   PlanMembership[]
```
kdata12, line 72: This should be planMembership (my mistake) haha

backend/prisma/schema.prisma
kdata12: We should use an enum instead of string here for better data integrity and to reduce errors

https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#defining-enums

backend/prisma/schema.prisma
fahadakhan2: done!

backend/prisma/schema.prisma
fahadakhan2: No worries! fixed

backend/prisma/schema.prisma
fahadakhan2: okay I think I added the enum correctly now

backend/prisma/schema.prisma
fahadakhan2: done!

backend/prisma/schema.prisma
fahadakhan2: Yeah that makes sense. Fixed!

backend/prisma/schema.prisma
```diff
@@ -118,23 +118,22 @@ model Contact {
...
 }
 
 model UserContactRelationship {
-  id               String  @id @default(uuid())
-  contact          Contact @relation(fields: [contactID], references: [id])
+  id               String           @id @default(uuid())
+  contact          Contact          @relation(fields: [contactID], references: [id])
   contactID        String
-  relationshipType String
+  relationshipType RelationshipType @default(OTHER)
```
kdata12: Looks good!

backend/prisma/schema.prisma
```diff
@@ -118,23 +118,22 @@ model Contact {
   owner              User                      @relation(fields: [ownerID], references: [id])
   importantDateEvent ImportantDateEvent[]
   relationships      UserContactRelationship[]
-  Plan               Plan?                     @relation(fields: [planId], references: [id])
-  planId             String?
```
kdata12: Here, we should still have `Plan  Plan?  @relation(fields: [planId], references: [id])` and  `planId   String?`

backend/prisma/schema.prisma
```diff
@@ -69,7 +70,9 @@ model Plan {
   specialEventType EventType
   createdAt        DateTime         @default(now())
   updatedAt        DateTime         @updatedAt
-  PlanMembership   PlanMembership[]
+  planMembership   PlanMembership[]
+  contacts         Contact[]
```
kdata12: If we're trying to establish that one plan can only be associated to one contact, this `contacts    Contact[]` should be removed. Instead we should have:

```js
model Plan {
   other fields...
   contact          Contact          @relation(fields: [contactID], references: [id])
   contactID        String
}

Comment (fahadakhan2) [reply backend/prisma/schema.prisma]: this makes sense done!

File: backend/prisma/schema.prisma (hunk 27 above)
Comment (fahadakhan2) [reply backend/prisma/schema.prisma]: We got this fixed so ima just ignore this

File: backend/prisma/schema.prisma (hunk 26 above)
Comment (fahadakhan2) [reply backend/prisma/schema.prisma]: Thank you sir!

Comment (fahadakhan2) [reply backend/prisma/schema.prisma]: Decided to not do the enums for this. Might come back and do this enum transformation later on but not a priority rn

File: backend/package.json (hunk 13 above)
Comment (fahadakhan2) [reply backend/package.json]: done!

File: frontend/src/components/PlanForm.tsx (hunk 4 above)
Comment (fahadakhan2) [reply frontend/src/components/PlanForm.tsx]: hyun's feauture that hes working on

File: frontend/src/components/PlanForm.tsx (hunk 5 above)
Comment (fahadakhan2) [reply frontend/src/components/PlanForm.tsx]: hyun's feauture

File: frontend/src/components/PlanForm.tsx (hunk 6 above)
Comment (fahadakhan2) [reply frontend/src/components/PlanForm.tsx]: hyuns feuture

File: frontend/src/components/SearchBarFilter.tsx
Comment (fahadakhan2) [reply frontend/src/components/SearchBarFilter.tsx:1]: hyun did this

File: frontend/src/components/contactsList.tsx (hunk 8 above)
Comment (fahadakhan2) [reply frontend/src/components/contactsList.tsx]: I added the route

File: backend/prisma/schema.prisma [hunk 29]
```diff
@@ -69,7 +70,40 @@ model Plan {
   specialEventType EventType
   createdAt        DateTime         @default(now())
   updatedAt        DateTime         @updatedAt
-  PlanMembership   PlanMembership[]
+  planMembership   PlanMembership[]
+  contact          Contact          @relation(fields: [contactID], references: [id])
+  contactID        String
```
kdata12: can you make this ContactID `String?` for not required.

fahadakhan2: done

.eslintrc.js
```diff
@@ -14,9 +14,11 @@ module.exports = {
     project: true,
     tsconfigRootDir: __dirname
   },
-
   rules: {
     eqeqeq: ['error', 'always'],
+    '@typescript-eslint/consistent-type-definitions': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/strict-boolean-expressions': 'off',
```
justinpchen94: actually i just saw this in hyun's pr and let's keep this rule intact - explicit null and explicit undefined checks are generally good things anyway

backend/app.ts
```diff
@@ -61,10 +62,12 @@ app.use(
 app.use(passport.session())
 
 passport.serializeUser((user: any, done: any) => {
+  console.log(user)
```
justinpchen94: rm

backend/app.ts
```diff
@@ -35,8 +36,8 @@ app.use(cors({
   origin: 'http://localhost:3000'
 }))
 
-app.use(express.json())
 app.use(passport.initialize())
+app.use(express.json())
```
justinpchen94: any reason the ordering was changed?

backend/app.ts
```diff
@@ -61,10 +62,12 @@ app.use(
 app.use(passport.session())
 
 passport.serializeUser((user: any, done: any) => {
+  console.log(user)
   return done(null, user.id)
 })
 
 passport.deserializeUser(async (id: string, done: any) => {
+  console.log(id)
```
justinpchen94: rm

backend/modules/auth.ts
```diff
@@ -1,7 +1,7 @@
 import { type NextFunction, type Request, type Response } from 'express'
 
 export const checkUserAuthorization = (req: Request, res: Response, next: NextFunction): void => {
-  if (req.user === null) {
+  if (!req.user) {
```
justinpchen94: we can revert this back after the explicit check rule is reverted

backend/prisma/schema.prisma
```diff
@@ -69,7 +70,40 @@ model Plan {
...
+
+model ImportantDateEvent {
+  id        String  @id @default(uuid())
+  contact   Contact @relation(fields: [contactID], references: [id])
+  contactID String
+  date      String
+  eventType String
+}
+
+model UserContactRelationship {
```
justinpchen94: isn't this model missing which user the contact has a relationship with?

backend/prisma/schema.prisma
```diff
@@ -102,6 +137,22 @@ enum InviteStatus {
   NOT_APPLICABLE
 }
 
+model GiftChoice {
```
justinpchen94: were these gift tables required in this pr? i think this was pulled over from hyun's pr right? if it's not directly relevant to your work let's rm

backend/routes/contactCreator.ts
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
+import { type User } from '@prisma/client'
+import express from 'express'
+import prisma from '../utils/prismaClient'
+
+const contactCreatorRouter = express.Router()
+const db = prisma
```
justinpchen94: does `import db from '../utils/prismaClient'` not work?

backend/routes/contactCreator.ts
```diff
@@ -0,0 +1,64 @@
...
+
+contactCreatorRouter.get('/', async function (req, res) {
+  try {
+    const contacts = await db.contact.findMany({
+      where: { ownerID: (req.user as User).id },
+      include: {
+        relationships: true,
+        importantDateEvent: true
+      }
+    })
```
justinpchen94: some good reading on some best practices with file structure:

https://blog.devgenius.io/the-three-layer-architecture-for-node-js-applications-ce32a3a30fa6

generally the data access layer should be in its own file and not called directly in the route handlers like we do here

see how hyun did it for Plan https://github.com/wewrap/togetherwewrap/blob/main/backend/routes/planForm.ts

Theres a route file, a controller file, and at least 1 model file that we want to split things up into

backend/routes/contactCreator.ts
```diff
@@ -0,0 +1,64 @@
...
+        lastName: req.body.lastName,
+        email: req.body.email,
+        phoneNumber: req.body.phoneNumber,
+        notes: req.body.notes,
+        ownerID: (req.user as User).id,
+        source: req.body.source
+      }
+    })
+    await db.userContactRelationship.createMany({
+      data: req.body.relationships.map((relationship: { relationshipType: any }) => ({
```
justinpchen94: let's try and provide more useful types than `any`

backend/routes/contactCreator.ts
```diff
@@ -0,0 +1,64 @@
...
+        relationshipType: relationship.relationshipType,
+        contactID: contact.id
+      }))
+    })
+    await db.importantDateEvent.createMany({
+      data: req.body.importantDates.map((importantDate: { date: any, eventType: any }) => ({
```
justinpchen94: same: avoid `any`

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
```
justinpchen94: let's avoid these file level ignores - either configure it in the eslint config or localize to only the offending line

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+    setImportantDates((prevDates) => [
+      ...prevDates.slice(0, index),
+      ...prevDates.slice(index + 1)
+    ])
+  }
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault()
+    try {
+      await axios.post('http://localhost:8000/contacts', {
```
justinpchen94: reminder that this url will not work once we deploy the app. add a todo here to change this url

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+              </label>
+
+              <label>
+                Event:
+                <input
+                  type="text"
+                  value={eventType}
+                  onChange={(event) => {
+                    const value = event.target.value
+                    if (/^[A-Za-z\s]*$/.test(value)) {
```
justinpchen94: let's pull this regex into a top level global variable

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+                />
+              </label>
+              <button type='button' onClick={addImportantEvent}>Add</button>
+              {importantDates.map((date, index) => (
+                <div key={index}>
+                  <input
+                    type="text"
+                    value={date.date}
+                    onChange={(event) => { handleImportantEventChange(index, 'date', event.target.value) }
+                    }
```
justinpchen94: maybe i can't match brackets correctly without an ide but this feels like it doesn't belong here?

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+                  />
+                  <input
+                    type="text"
+                    value={date.eventType}
+                    onChange={(event) => { handleImportantEventChange(index, 'eventType', event.target.value) }
+                    }
```
justinpchen94: same here potentially?

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+                  />
+                  <button type="button" onClick={() => { handleRemoveImportantEvent(index) }}>
+                    x
+                  </button>
+                </div>
+              ))}
+            </div>
+            <div>
+              <button className="add_contact_button" type="submit">
```
justinpchen94: i think we're using camelCase in css already so let's keep it consistent one way or the other

frontend/src/components/contactsList.tsx
justinpchen94: ah i misread this earlier - i don't know if the catch in the getUserData gets thrown correctly if this line 51 here doesn't also include a .catch block

fahadakhan2: Is this what you mean by including another catch block. Sorry if I'm not understanding this correctly.
 useEffect(() => {
    const getContacts = async () => {
      try {
        const response = await axios.get('http://localhost:8000/contacts', { withCredentials: true })
        const contactsData = response.data.contacts as Contact[]
        setContacts(contactsData)
      } catch (error) {
        console.error(error)
      }
    }
    try {
      void getContacts()
    } catch (error) {
      console.error(error)
    }
  }, [])

justinpchen94: i think you can do it as 

```
getContacts().catch()

```

.eslintrc.js
fahadakhan2: sounds good!

backend/app.ts
fahadakhan2: Oh we did this when trying to debug I can put it back haha

backend/modules/auth.ts
fahadakhan2: done!

backend/prisma/schema.prisma
fahadakhan2: Okay I will remove anything thats not related to user and its contacts

backend/routes/contactCreator.ts
fahadakhan2: yeah this worked thanks!

backend/routes/contactCreator.ts
fahadakhan2: sounds good!

backend/routes/contactCreator.ts
fahadakhan2: changed to string

frontend/src/components/contactsForm.tsx
fahadakhan2: removed

frontend/src/components/contactsForm.tsx
fahadakhan2: done!

frontend/src/components/contactsForm.tsx
fahadakhan2: It does belong but Ill just move it to one line so it doesn't take up more space than necessary.

frontend/src/components/contactsList.tsx
fahadakhan2: this gives me the err: Expected an assignment or function call and instead saw an expression.

But this ends up working: getContacts().catch(console.error)

lmk if that works.

backend/prisma/schema.prisma
kdata12: This should be renamed to `ContactRelationship`, i think the name is misleading. This model exists to store each contact relationships
image

fahadakhan2: Changed the name

backend/app.ts
```diff
@@ -175,6 +176,7 @@ app.use('/auth/facebook', facebookOAuthRouter)
 app.use('/login', loginAuthRouter)
 app.use('/api/plan', checkUserAuthorization, planFormRouter)
 app.use('/signup', signUpAuth)
+app.use('/contacts', contactCreatorRouter)
```
justinpchen94: nit: this should be similar to `api/plan` above

backend/controllers/contacts.ts
```diff
@@ -0,0 +1,53 @@
...
+      const contactData = {
+        firstName,
+        lastName,
+        email,
+        phoneNumber,
+        notes,
+        source,
+        ownerID,
+        relationships: req.body.relationships,
+        importantDateEvent: req.body.importantDateEvent
```
justinpchen94: it doesn't seem worth it to pull the individual fields off of the object in line 8-14 if you're going to reference them directly as well on line 25-26. picking them all out seems unnecessary given you don't actually use the variables anywhere else, so i'd just remove lines 7-15 and then update the current line 18-24 to just read off of the req.body directly

```
const contactData = {
  firstName: req.body.firstName,
  ...
}

```

backend/controllers/contacts.ts
```diff
@@ -0,0 +1,53 @@
...
+      res.status(200).json(contacts);
+    } catch (err) {
+      console.error('error in get contacts', err);
+      res.status(400).send('error');
+    }
+  }
+}
+
+export const createContact = ContactController.createContact
+export const getContacts = ContactController.getContacts
```
justinpchen94: prob just remove these, since exporting functions like this sort of defeats part of the purpose of declaring a public class like you do on line 4


```
export default class ContactController {
```

you'd normally import and call functions like:
```
import ContactController from 'backend/controllers/contacts';

...
ContactController.getContacts(...);
```

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+    if (relationshipType.trim() === '') {
+      setShowError(true)
+      return
+    }
+    setRelationships((prevRelationships) => [
+      ...prevRelationships,
+      {
+        relationshipType
+      }
+    ])
```
justinpchen94: is this not the same thing as just 

```
setRelationships(relationships.concat([{ relationshipType }])
```

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+  const handleRelationshipChange = (
+    index: number,
+    field: keyof Relationship,
+    value: any
+  ) => {
+    setRelationships((prevRelationships) =>
+      prevRelationships.map((type, i) =>
+        i === index ? { ...type, [field]: value } : type
+      )
+    )
```
justinpchen94: i believe you can simplify this using the array [splice method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  const handleRemoveRelationshipType = (index: number) => {
+    setRelationships((prevRelationships) => [
+      ...prevRelationships.slice(0, index),
+      ...prevRelationships.slice(index + 1)
```
justinpchen94: same, this is also doable with `splice`

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+    if (eventDate.trim() === '' || eventType.trim() === '') {
+      setShowError(true)
+      return
+    }
+    setImportantDates((prevDates) => [
+      ...prevDates,
+      {
+        date: eventDate,
+        eventType
+      }
```
justinpchen94:
```
setImportantDates(importantDates.concat([{ date: eventDate, eventType }]))
```

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+
+  const handleImportantEventChange = (
+    index: number,
+    field: keyof ImportantDate,
+    value: any
+  ) => {
+    setImportantDates((prevDates) =>
+      prevDates.map((date, i) =>
+        i === index ? { ...date, [field]: value } : date
+      )
```
justinpchen94: splice

frontend/src/components/contactsList.tsx
justinpchen94: are you going to add error states later? this simply logs an error to the console (which may not always be what we want to do)

backend/prisma/migrations/20230421054329_new_migration/migration.sql
justinpchen94, line 1: in the future let's use more useful migration names than `new_migration` or `updated_migration`

frontend/src/App.tsx
```diff
@@ -15,7 +15,7 @@ function App (): JSX.Element {
         <Route path="/login" element={<LoginForm />} />
         <Route path="/signup" element={<SignUp />} />
         <Route path="/tempLandingPage" element={<TempLandingPage />} />
-        <Route path="/contactPage" element={<Contacts />} />
+        <Route path="/contactPage" element={<ContactsList />} />
```
justinpchen94: nit: the url doesn't seem right to me if it's called `wewrap.com/contactPage` I think it'd normally just be `wewrap.com/contacts`

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+type Relationship = {
+  relationshipType: string
+}
+
+interface Props {
+  setShowCreateAContactForm: (value: boolean) => void
+  handleContactCreate: (newContact: any) => void
+}
+
+export const CreateAContactForm = ({ setShowCreateAContactForm, handleContactCreate }: Props) => {
```
justinpchen94: it's not normal to add "A" into variables names

`createContactForm`

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+              <label htmlFor="first_name">First name</label>
+              <input
+                className="first_name"
+                name="first_name"
+                type="text"
+                autoComplete="on"
+                required
+                value={firstName}
+                onChange={handleFirstNameChange}
+              ></input>
```
justinpchen94: if you don't have any child elements to nest, I think this is just `<input ... />` and you don't need a pair of tags 


```
              <input
                className="first_name"
                name="first_name"
                type="text"
                autoComplete="on"
                required
                value={firstName}
                onChange={handleFirstNameChange}
              />
```

frontend/src/components/contactsForm.tsx
```diff
@@ -0,0 +1,310 @@
...
+                name="notes"
+                type="text"
+                autoComplete="on"
+                required
+                value={notes}
+                onChange={handleNotesChange}
+              ></input>
+
+              <label>
+                Relationship with Person
```
justinpchen94: nit: should this just be **Relationship with Contact** instead of person?

frontend/src/components/contactsList.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useEffect, useState } from 'react'
+import { CreateAContactForm } from './contactsForm'
+import axios from 'axios'
+
+interface Contact {
```
justinpchen94: is this copied directly from prisma? shouldn't we just import `Contact` from prisma instead?

backend/app.ts
fahadakhan2: done!

backend/controllers/contacts.ts
fahadakhan2: Makes sense. Changed!

backend/controllers/contacts.ts
fahadakhan2: Done!

frontend/src/App.tsx
fahadakhan2: Yeah I agree will change!

frontend/src/components/contactsForm.tsx
fahadakhan2: Done!

package-lock.json
justinpchen94, line 1: maybe remove this whole file?

frontend/src/components/contactsList.tsx
```diff
@@ -0,0 +1,90 @@
...
+export const ContactsList = () => {
+  const [showCreateAContactForm, setShowCreateAContactForm] = useState<boolean>(false)
+  const [contacts, setContacts] = useState<Contact[]>([])
+
+  const toggleContactForm = () => {
+    setShowCreateAContactForm(!showCreateAContactForm)
+  }
+
+  const handleContactCreate = (newContact: Contact) => {
+    // setContacts((prevState) => [...prevState, newContact])
```
justinpchen94: i think this comment is the code you wanted right? rm the line 39 code instead?

justinpchen94 (approved):

frontend/src/components/contactsList.tsx
fahadakhan2: Done!

PR #35, opened 2023-03-08
T90: Update data model

kdata12:
**Added relation models to `schema.prisma`**

Updated Diagram 3/20:
image


backend/prisma/schema.prisma
```diff
@@ -8,20 +8,38 @@ datasource db {
...
 
 model User {
-  id           String    @id @default(uuid())
-  email        String   @unique
-  firstName    String
-  lastName     String
-  password     String?
-  birthDate    DateTime?
-  googleID     String?   @unique
+  UserID      String             @id @default(uuid())
```
justinpchen94: generally it's normal to leave the id field of the table to simply be `id`, and it's somewhat a convention that when it's a FK to name it based on the table

backend/prisma/schema.prisma
```diff
@@ -8,20 +8,38 @@ datasource db {
...
+  friendsWith        User               @relation("friendsWith", fields: [friendsWithID], references: [UserID])
+  friendsWithID      String
+  relationshipStatus RelationshipStatus @default(NOT_FRIEND)
+}
 
+enum RelationshipStatus {
+  FRIEND
+  PENDING_REQUEST
+  NOT_FRIEND
 }
```
justinpchen94, line 46: while it is legal syntax to leave enum like this, it sets the enum value to be numeric.

`RelationshipStatus.FRIEND` will have a value of `1`
which is why we generally try to do 

`FRIEND = 'friend'`

https://www.typescriptlang.org/docs/handbook/enums.html

backend/utils/seed.ts
```diff
@@ -0,0 +1,148 @@
...
+            role: 'PLAN_LEADER'
+        }
+    })
+
+    //hyun sends john as a friend first
+    const hyunAddsJohn = await db.userRelationship.create({
+        data: {
+            userID: hyun.UserID,
+            friendsWithID: john.UserID,
+            relationshipStatus: "PENDING_REQUEST"
```
justinpchen94: let's avoid these magic strings! you already defined an enum, so let's use that here

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+
+model Pledge {
+  PledgeID     String        @id @default(uuid())
+  plan         Plan          @relation(fields: [planID], references: [PlanID])
+  planID       String
+  user         User          @relation(fields: [userID], references: [UserID])
+  userID       String
+  inviteStatus InviteStatus?
+  role         Role
+  giftChoice   GiftChoice[]
```
justinpchen94: should a pledge be tied to a gift or to a plan?

backend/prisma/schema.prisma
justinpchen94: same with all the other enums below

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
 }
 
 model Profile {
-  id     String @id @default(uuid())
-  user   User @relation(fields: [userId], references: [id])
-  userId String @unique @default(uuid())
-  photoUrl String? 
-}
\ No newline at end of file
+  profileID String  @id @default(uuid())
```
justinpchen94: should be `id`

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  user      User    @relation(fields: [userId], references: [UserID])
+  userId    String  @unique
+  photoUrl  String?
+}
+
+model Plan {
+  PlanID           String      @id @default(uuid())
```
justinpchen94: `id`

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  createdDate      DateTime    @default(now())
+  specialEventType EventType
+  contact          Contact?    @relation(fields: [contactID], references: [ContactID])
+  contactID        String?     @unique
+  User             Pledge[]
+  FinalGift        FinalGift?
+}
+
+model Pledge {
+  PledgeID     String        @id @default(uuid())
```
justinpchen94: id

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+
+model Plan {
+  PlanID           String      @id @default(uuid())
+  description      String
+  startDate        DateTime
+  endDate          DateTime
+  createdDate      DateTime    @default(now())
+  specialEventType EventType
+  contact          Contact?    @relation(fields: [contactID], references: [ContactID])
+  contactID        String?     @unique
```
justinpchen94: let's be consistent with the suffix `ID` vs `Id` if possible

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  User             Pledge[]
+  FinalGift        FinalGift?
+}
+
+model Pledge {
```
justinpchen94, line 75: think we're still missing
* amount
* pledgeStatus (since inviteStatus doesn't seem to correlate specifically to the state of the monetary pledge)

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+enum InviteStatus {
+  INVITED
+  RESENT_INVITE
+  ACCEPTED
+  DENY
+  NOT_APPLICABLE
+}
+
+model Contact {
+  ContactID          String                    @id @default(uuid())
```
justinpchen94: `id`

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  email              String?
+  firstName          String
+  lastName           String?
+  phoneNumber        String?
+  address            String
```
justinpchen94: maybe we omit this for now since fahad is punting on address for now - he should be migrating this field onto the Contact table separately once he figures out the implementation there

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  source             String
+  owner              User                      @relation(fields: [ownerID], references: [UserID])
+  ownerID            String
+  importantDateEvent ImportantDateEvent[]
+  relationship       UserContactRelationship[]
+  Plan               Plan?
+}
+
+model ImportantDateEvent {
+  ConImportEventID String  @id @default(uuid())
```
justinpchen94: `id`

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  contactID        String
+}
+
+model UserContactRelationship {
+  UserContactRelationshipID String  @id @default(uuid())
+  contact                   Contact @relation(fields: [contactID], references: [ContactID])
+  contactID                 String
+}
+
+model GiftChoice {
```
justinpchen94: i think this can just be a `Gift`. Jess will be owning the gift voting and etc so all we're responsible for in this PR right now is to create a table representing all possible gifts, and then be able to link that to a Plan, so we probably don't need a Final Gift table

justinpchen94: Gifts should live separate from the voting/choice mechanism cuz Gifts will also be used as part of wishlists and marketplaces and etc

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  GiftChoiceID String   @id @default(uuid())
```
justinpchen94: `id`

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  Pledge       Pledge   @relation(fields: [pledgeID], references: [PledgeID])
+  pledgeID     String
```
justinpchen94: remove for now since we're adapting this to a `Gift` table

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  giftUrl      String
+  giftName     String
```
justinpchen94: the `gift` prefix is somewhat assumed so we can just name these `url` and `name`

backend/prisma/schema.prisma
```diff
@@ -32,8 +50,100 @@ model Session {
...
+  dateCreated  DateTime @default(now())
```
justinpchen94: general convention is to add `createdAt` and `updatedAt` for all models so let's rename and add that in every model too

https://stackoverflow.com/questions/65474757/nexus-prisma-how-to-handle-createdat-and-updatedat-with-crud-globally

backend/utils/seed.ts
```diff
@@ -0,0 +1,148 @@
+import { faker } from '@faker-js/faker'
```
justinpchen94: let's name this seed script to be a little more specific to what data in particular is being seeded here

justinpchen94: also i'd encourage you to be conscientious about whether we should use a random set of values every time or a consistent set of values and what the pros and cons of each approach would be

backend/utils/seed.ts
```diff
@@ -0,0 +1,148 @@
...
+    lastName: faker.name.lastName(),
+    email: faker.internet.email()
+}))
+
+
+const main = async () => {
+
+    //create many users
+    await db.user.createMany({
+        data: userData
```
justinpchen94: given this is a small script, performance isn't super important but you could just do all of these calls at once


```
await db.user.createMany({
    data: [...userdata, {
        /* hyun user */ 
    }, { 
        /* john user */
    }, {
        /* sarah user */
    }]
})

```

backend/utils/seed.ts
```diff
@@ -0,0 +1,148 @@
...
+            relationshipStatus: "FRIEND"
+        }
+    })
+}
+
+main()
+    .then(
+        async () => {
+            await db.$disconnect()
+            console.log('Disconnected fom DB')
```
justinpchen94: coudl also make this message more useful and say something like 'script executed successfully' or something

backend/utils/seed.ts
```diff
@@ -0,0 +1,148 @@
...
+        }
+    )
+    .catch(async (e) => {
+        console.error(e)
```
justinpchen94: same here, can prefix the error with more info

justinpchen94 (commented):
ill allow it for now but remember we will eventually get to a point where we _should not_ be deleting migrations unless absolutely necessary. the staff will probably start enforcing this more heavily soon and will definitely disallow this from being a continued pattern by the time we start deploying things to a live db

backend/prisma/schema.prisma
kdata12: Oh this is not the enum from typescript. It's prisma way of defining enum for the ENUM constraint.
https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#defining-enums

justinpchen94: ah in that case, let's still make sure these enum's don't drift from the typescript ones we'll have to use as well

https://github.com/prisma/prisma/discussions/9215

backend/utils/seed.ts
kdata12: The prisma client display the enum list when I enter an empty string. 
image

Do you want me to use the typescript enum here

justinpchen94: yeah using an enum value in general would be preferred to just having these strings float around for just a bunch of reasons

kdata12: Imported prisma enum 
image

image

justinpchen94: sick!

backend/prisma/schema.prisma
kdata12: My reasoning is that GiftID can be uniquely identify by the pledgeID, and in turn, the pledgeID is uniquely defined by the planID and userID. I think gift choices, like casting vote for a gift in a plan, should be stored with the pledgeID and not the planID.

If we do decide store the gift choices in the plan table, we'd have to make another join table between the Plan and User called 'UserGiftChoice' which is essentially the 'pledge' table rn.

justinpchen94: not sure if im just reading your comment wrong but the way I'm thinking about this is essentially that there is a 1 to many relationship between Plans and Pledges, or that a plan can have many pledges, but a pledge can have only one plan. this is represented correctly in your schema I believe

the question I'm asking essentially is should the pledge _have_ to be tied to a gift? consider the following sequence of events:
1) Plan leader chooses a laptop as the gift
2) I pledge $50 towards the laptop and submit $50
3) plan leader decides to switch to a ps5 instead

do we have to update the giftchoice for every pledge every time the gift is changed? that seems like unnecessary extra work to me, when we could just link a pledge strictly to a Plan and then not worry ourselves about the gift, since the Plan acts as the central linking entity

kdata12: Added gift choice table

backend/utils/seed.ts
kdata12: I was thinking about this, I try doing a mix of both. I created 3 consistent users and populate them with fake contacts for now.

backend/utils/seed.ts
kdata12: I created the variable for each users (hyun, sarah, and john) so I can conveniently access their properties for other queries.

backend/prisma/schema.prisma
```diff
@@ -32,8 +54,111 @@ model Session {
...
+  endDate          DateTime
+  createdDate      DateTime     @default(now())
+  specialEventType EventType
+  contact          Contact?     @relation(fields: [contactID], references: [id])
+  contactID        String?      @unique
+  User             Pledge[]
+  createdAt        DateTime     @default(now())
+  updatedAt        DateTime     @updatedAt
+  GiftChoice       GiftChoice[]
+}
```
chinkevin25: Do we need a `createdBy`? 
we also have a `createdAt` and a `createdDate` field. I believe we only need one.

backend/prisma/schema.prisma
chinkevin25: It does make me wonder though what actually gets saved into the database? I feel like i'd be a number still right? I think it's fine to leave it this way (though I would recommend finding a way to maintain a single source of truth if possible). Do you know the answer?

backend/prisma/schema.prisma
```diff
@@ -32,8 +54,111 @@ model Session {
...
+  RESENT_INVITE
+  ACCEPTED
+  DENY
+  NOT_APPLICABLE
+}
+
+model Contact {
+  id                 String                    @id @default(uuid())
+  email              String?
+  firstName          String
```
chinkevin25: Do we not need last name?

backend/prisma/schema.prisma
```diff
@@ -32,8 +54,111 @@ model Session {
...
+  notes              String?
+  source             String
+  owner              User                      @relation(fields: [ownerID], references: [id])
+  ownerID            String
+  importantDateEvent ImportantDateEvent[]
+  relationship       UserContactRelationship[]
+  Plan               Plan?
```
chinkevin25: I'm not super well versed in Prisma so i'd defer to @justinpchen94 here. 

Don't we want to save the plan ID here and shouldn't this be a 1 Contact to Many Plan relationship too? Each contact will eventually have more than one plan associated with them (i.e. their second birthday on the system).

How are we modeling that right now?

backend/prisma/schema.prisma
```diff
@@ -32,8 +54,111 @@ model Session {
...
+  price     Int
+}
+
+
+enum EventType {
+  BIRTHDAY
+  ANNIVERSARY
+  GRADUATION
+  WEDDING
+  BABYSHOWER
```
chinkevin25: `BABY_SHOWER` i assume since it's 2 words?

backend/utils/seed.ts
```diff
@@ -0,0 +1,146 @@
+//seeds fake users and contacts into the database.
+import { faker } from '@faker-js/faker'
```
chinkevin25: Not something I'm worried about, but we'll probably want to not include this file (and faker js) in our production build (not that it really matters but good etiquette might suggest we do that). We can tackle that when we discuss production builds

backend/prisma/schema.prisma
kdata12: For the `createdBy` user, we're storing the user who created the plan in another join table called `Pledge` that joins together the `User` and `Plan` table!

Ok I deleted createdDate, good catch!

backend/prisma/schema.prisma
kdata12: Yeah it does show up in the DB as `FRIEND`

image

backend/prisma/schema.prisma
kdata12: Justin told me to leave this table for Fahad to figure it out 🙂

chinkevin25: fosho

backend/prisma/schema.prisma
kdata12: What I had in mind was that only 1 contact can 0 or 1 plan. We can do 1 Contact to Many Plan.

kdata12: image

justinpchen94 (approved):

PR #36, opened 2023-03-13
T69: Build Landing Page

midwestbread:


frontend/public/landingfolder/landingpage.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
```
justinpchen94, line 1: let's try and get this built in react! this is how the HTML should end up rendering but let's build this out as a react component/page

PR #37, opened 2023-03-13
T91: Plan Form Frontend

kdata12:
Added plan form to the frontend. Created react components for plan form.

For the eslint: 
- ~I went the hacky way to configure the eslint. I didn't want to spend too time on it.~
- I moved eslint to ./frontend (this is probably wrong) just to make it work with react.

image


frontend/package.json
```diff
@@ -24,10 +24,8 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
-  },
-  "eslintConfig": {
-    "extends": []
+    "eject": "react-scripts eject",
+    "lint": "eslint src/ --ext .tsx"
```
justinpchen94, line 28: this doesn't lint everything right? should we update this command?

frontend/src/App.tsx
```diff
@@ -1,26 +1,24 @@
...
-    </Routes>
-</div>
-  );
+    <div>
+      <Routes>
+        <Route path="/login" element={<LoginForm />} />
+        <Route path="/signup" element={<SignUp />} />
+        <Route path="/tempLandingPage" element={<TempLandingPage />} />
+        <Route path="/contactPage" element={<Contacts />} />
+        <Route path="/planForm" element={<PlanForm />} />
```
justinpchen94, line 18: not necessary in this pr but think about how you would protect certain pages behind the auth wall. how could we prevent logged out users from accessing logged in pages?

frontend/src/App.tsx
```diff
@@ -1,26 +1,24 @@
...
+      </Routes>
+    </div>
+  )
 }
 
-export default App;
```
justinpchen94, line 26: maybe it's just me being old school but I'm team require semicolon. if you are too, might be worth updating the eslint config and adding a rule override that restores the semicolon rule

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,236 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
```
justinpchen94: this is generally a sign that you can probably just disable the rule entirely if you don't agree with it, rather than just ignoring whole files

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+import React, { useState } from 'react'
+import { SearchBarFilter } from './SearchBarFilter'
```
justinpchen94, line 2: not important but i like to think about my import orders as usually having package imports (like 'react') at the top, and then the project specific imports second

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>): Promise<void> => {
+    event.preventDefault()
+    if (specialPerson === undefined) {
+      handleError('Please add 1 special person')
+      return
+    }
+    try {
+      await axios.post('http://localhost:8000/planform', {
```
justinpchen94: just remember we'll want to remove the hardcoded source url's eventually!

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+    setErrorMessage(message)
+    setTimeout(() => {
+      setErrorMessage('')
+      setError(false)
+    }, 4000)
+  }
+
+  const handleEventSelect = (e: React.ChangeEvent<HTMLSelectElement>): void => {
+    setEventType(e.target.value as EventType)
+  }
```
justinpchen94: js has something called "hoisting" which you are taking advantage of (knowingly or not) here, but generally it's still good practice to define functions before they are referenced, so let's move this up to where the other `handle` functions live

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+        friends,
+        eventType
+      }, {
+        withCredentials: true
+      })
+    } catch (error) {
+      if (error instanceof AxiosError) {
+        console.error(error?.response?.status)
+        console.error(error?.response?.data)
+      }
```
justinpchen94: anything else you want to do if the post request doesn't work?

frontend/src/components/PlanForm.tsx
justinpchen94: also let's camel case the route correctly - it should be `/planForm`

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+    }
+  }
+
+  const handleError = (message: string): void => {
+    setError(true)
+    setErrorMessage(message)
+    setTimeout(() => {
+      setErrorMessage('')
+      setError(false)
+    }, 4000)
```
justinpchen94: haha clever way of temporarily showing the error!

there can be a race condition here though where someone submits the form, encounters an error, then re submits again and generates a diff error, and this setTimeout would fire and clear the error that is showing to the user the 2nd submit

frontend/src/components/PlanForm.tsx
```diff
@@ -0,0 +1,236 @@
...
+        </div>
+        <div>
+          <button type='submit'>Create Plan</button>
+        </div>
+      </form>
+    </div>
+  )
+}
+
+const fakeUserData = [
```
justinpchen94: again, let's define this earlier in the file

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable array-callback-return */
+/* eslint-disable @typescript-eslint/strict-boolean-expressions */
```
justinpchen94: same thing - look into overriding the rule entirely

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
...
+import React, { useState } from 'react'
+import { Tag } from './Tag'
+import './searchbar.css'
+import { type Friend } from './PlanForm'
+
+interface Props {
+  items: Friend[]
+  handleSelectChange: (friend: Friend) => void
+  handleRemoveTag?: (friend: Friend[]) => void
+  numbOfSelect?: number
```
justinpchen94: nit: when shortening `number` it's more common to use `num` 😂

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable array-callback-return */
+/* eslint-disable @typescript-eslint/strict-boolean-expressions */
+import React, { useState } from 'react'
+import { Tag } from './Tag'
+import './searchbar.css'
+import { type Friend } from './PlanForm'
```
justinpchen94, line 5: this is interesting syntax to me since I normally see it as `import type {Friend} from '...'`

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export const SearchBarFilter: React.FC<Props> = ({
+  items,
+  handleSelectChange,
+  handleRemoveTag,
+  numbOfSelect,
+  setSelectError
+}) => {
+  const [query, setQuery] = useState<string>()
```
justinpchen94: no default here?

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
...
+    setQuery(value)
+
+    const filteredItems: Friend[] = items.filter(
+      obj => obj.firstName.toLowerCase().includes(value)
+    )
+    setItems(filteredItems)
+  }
+
+  const handleSelect = (friend: Friend): void => {
+    if (numbOfSelect &&
```
justinpchen94: generally im a little wary of using the js truthy shorthand, since expressions like `if (numbOfSelect) {}` may not evaluate the way you think it does for numbers and strings. `0` and `''` empty string both evaluate to false, which may be incorrect behavior if your intent is to check "is this variable defined"/"is this variable null or undefined"

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
...
+      select.length < numbOfSelect &&
+      !select.includes(friend)) {
```
justinpchen94: is this condition possible to be false? specifically I mean the `select.includes(friend)`?

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
...
+      handleSelectChange(friend)
+      setSelected(select => [...select, friend])
+    } else if (!numbOfSelect && !select.includes(friend)) {
+      handleSelectChange(friend)
+      setSelected(select => [...select, friend])
```
justinpchen94: these branches both do the same thing - any way to logically combine them?

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,87 @@
...
+              onClick={() => {
+                handleSelect(obj)
+              }}
+              key={obj.firstName}>{obj.firstName}
+            </div>))
+        )}
+      </div>
+      <p>Selected:</p>
+      <div className='tag-grid'>
+      {select.map(obj => (
```
justinpchen94: let's try and use descriptive variables everywhere if possible. `obj` isn't super helpful here, even if you're not doing anything complicated with it, it's still worth naming well!

frontend/src/components/loginForm.tsx
```diff
@@ -1,78 +1,79 @@
-import React, { useState } from 'react';
-import axios from 'axios';
-import './login.css';
+/* eslint-disable @typescript-eslint/no-misused-promises */
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
justinpchen94: worth updating rules!

frontend/src/components/Tag.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <div className='tag-wrapper' style={tagColor}>
+        <div className='tag-text'>{friend.firstName}</div>
+        <button className='tag-close-button' onClick={handleClick}> x </button>
+      </div>
+    </div>
+  )
+}
+
+// generate light hexademical color string
+function generateRandomLightColor (): string {
```
justinpchen94: a rare case of a function being named well enough to render the comment redundant! let's rm comment

frontend/src/components/Tag.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { type Friend } from './PlanForm'
+import './tag.css'
+
+interface Props {
+  friend: Friend
+  handleRemoveTag: (friend: Friend) => void
+}
+
+export const Tag: React.FC<Props> = ({ friend, handleRemoveTag }) => {
+  const [backgroundColor] = useState<any>(generateRandomLightColor())
```
justinpchen94: does this need to be a state variable? you can simply just declare this variable and use it directly without setting state

frontend/src/components/Tag.tsx
justinpchen94: also generally good to pull this logic out into a helper or util file instead

frontend/src/components/Tag.tsx
```diff
@@ -0,0 +1,37 @@
...
+  )
+}
+
+// generate light hexademical color string
+function generateRandomLightColor (): string {
+  const r = Math.floor(Math.random() * 100) + 155
+  const g = Math.floor(Math.random() * 100) + 155
+  const b = Math.floor(Math.random() * 100) + 155
+
+  return `#${r.toString(16)}${g.toString(16)}${b.toString(16)}`
```
justinpchen94: also this whole fn can be rewritten as:

```

'#' +
  ['r', 'g', 'b']
    .map((_c) => Math.floor(Math.random() * 100 + 155).toString(16))
    .join('');
```

frontend/src/components/signup.tsx
```diff
@@ -1,77 +1,73 @@
...
-                setWarning('Please use a different password.')
-            }
+        .catch((err) => {
+          if (err.response.status === 409) {
+            setWarning('Existing email. Please use a different email.')
+          } else { setWarning('Submission Error. Please try again.') }
+        })
+    } else {
+      setConfirmPassword('')
+      setWarning('Please use a different password.')
```
justinpchen94, line 38: shouldn't the warning here be more along the lines of "your passwords do not match"

frontend/src/App.tsx
kdata12: I'm currently on this in another branch, we would check if the user is logged in with useContext, if user is not logged in then they won't be able to access certain routes. I would have to consult with you and kevin to implement this correctly.

frontend/src/components/PlanForm.tsx
kdata12: Ohhh display an error message on UI on why the submission didn't work.

frontend/src/components/PlanForm.tsx
kdata12: Ahhh I see, I'll just put an `if (errorMessage.length > 0) return` to prevent the setTimeout from getting queued up and clearing the next message.

frontend/src/components/SearchBarFilter.tsx
kdata12: yeah I have to put the `type` there or else eslint will yell at me like my parents do (childhood trauma).

frontend/src/components/SearchBarFilter.tsx
kdata12: 😂

frontend/src/components/SearchBarFilter.tsx
kdata12: added default string to query

frontend/src/components/SearchBarFilter.tsx
kdata12: ok I just made `numOfSelect` a required prop to prevent this

frontend/src/components/SearchBarFilter.tsx
kdata12: done, renamed to friend

frontend/src/components/SearchBarFilter.tsx
kdata12: Yes, it can be false when you keep on trying to add the same person.

frontend/src/components/signup.tsx
kdata12: This is Jess's file, I just reformatted because of eslint was screaming

frontend/src/components/Tag.tsx
kdata12: Oh this scenario was pretty interesting. Let say I use `const backgroundColor = generateRandomLightColor()` then every time I render the `Tag` component, it will also regenerate a new color in the UI. But I use `useState`, the color will not regenerate when the `Tag` component re-renders.

frontend/src/components/Tag.tsx
kdata12: 🥶

frontend/src/App.tsx
justinpchen94: sounds good, just wanted to make sure you think about it

frontend/src/components/PlanForm.tsx
justinpchen94: yeaup!

backend/app.ts
```diff
@@ -171,4 +166,11 @@ passport.use(
         }
     ));
 
+app.use('/', testRouter)
+app.use('/auth/google', googleOAuthRouter)
+app.use('/auth/facebook', facebookOAuthRouter)
+app.use('/login', loginAuthRouter)
+app.use('/signup', signUpAuth)
+
+
```
justinpchen94: nit: no more than 1 newline in a row pls

backend/routes/loginAuth.ts
```diff
@@ -11,6 +11,12 @@ loginAuthRouter.get('/password',
     res.render('login');
   });
 
+loginAuthRouter.post("/password", passport.authenticate('local'),
+  function (req, res) {
+    console.log('user from local auth: ' + (req.user as User).id);
```
justinpchen94: as a side note we generally want to use `console.info` for stuff that you want to log that is relevant to log (aka, not just for debugging)

some background reading: https://sematext.com/blog/logging-levels/

backend/routes/testRoute.ts
```diff
@@ -1,3 +1,4 @@
+import { User } from "@prisma/client";
```
justinpchen94: doesn't look necessary in this file?

frontend/package.json
justinpchen94: look into also the `--fix` flag in eslint - might make everyone's life easier 😅

frontend/src/components/PlanForm.css
```diff
@@ -0,0 +1,15 @@
...
+  gap: 2em;
+}
+
+.error-message {
+  flex: 0 0 100%;
+  flex-wrap: wrap;
+  color: red;
+  padding: 0;
+  margin: 0;
+}
```
justinpchen94: generally good practice to leave a newline at end of files, which is what this red icon denotes

frontend/src/components/Tag.tsx
```diff
@@ -0,0 +1,30 @@
...
+import './tag.css'
+
+interface Props {
+  friend: Friend
+  handleRemoveTag: (friend: Friend) => void
+}
+
+export const Tag: React.FC<Props> = ({ friend, handleRemoveTag }) => {
+  const [backgroundColor] = useState<any>(generateRandomLightColor())
+  // const backgroundColor = generateRandomLightColor()
```
justinpchen94: rm comment

frontend/src/utils/colorGenerator.ts
```diff
@@ -0,0 +1,7 @@
+export function generateRandomLightColor (): string {
```
justinpchen94: nit: let's add something like "hex" in this fn name to make it more distinct from if this was generating, say RGBA values

`generateRandomLightColorHex`

justinpchen94 (approved):
mostly small nits, looking good!

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,82 @@
...
+export const SearchBarFilter: React.FC<Props> = ({
+  friendArray,
+  handleSelectChange,
+  handleRemoveTag,
+  numOfSelect,
+  setSelectError
+}) => {
+  const [query, setQuery] = useState<string>('')
+  const [friends, setFriends] = useState<Friend[]>([])
+  const [select, setSelected] = useState<Friend[]>([])
```
justinpchen94: use a better name here like `selectedFriend`

frontend/src/components/SearchBarFilter.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const handleQuery = (event: React.ChangeEvent<HTMLInputElement>): void => {
+    const value = event.target.value
+    setQuery(value)
+
+    const filteredItems: Friend[] = friendArray.filter(
```
justinpchen94: let's use "friends" instead of "items" here imo

PR #39, opened 2023-03-17
T124: Fix local auth cookie bug

kdata12:
## Description 

Fix passport local auth bug where user login did not create a cookie session.

Resource: https://github.com/jaredhanson/passport-local

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [] I have made sure that eslint was passed locally
- [] I have added necessary documentation (if appropriate)


backend/app.ts
```diff
@@ -27,9 +27,14 @@ const THREE_DAYS = 1000 * 60 * 60 * 24 * 3
 const TWO_MINUTES = 1000 * 60 * 2
 
 app.use(morgan("dev"));
-app.use(cors());
+
+app.use(cors({
+    credentials: true,
+    origin: 'http://localhost:3000'
```
justinpchen94, line 33: do we know yet what this will have to be once we deploy the app? should this instead be an env var?

justinpchen94 (approved):
small comment but lgtm to unblock

PR #42, opened 2023-03-30
T122: Plan form backend

kdata12:
## Description 

Created an api endpoint for handling plan form submission.

## Types of changes

## Checklist


kdata12:
You can ignore the frontend files, I just checked them out from `hyun/plan-form-frontend` to test the route!

backend/routes/planForm.ts
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable @typescript-eslint/no-misused-promises */
```
justinpchen94, line 1: let's make this a rule!

backend/routes/planForm.ts
```diff
@@ -0,0 +1,60 @@
...
+const planFormRouter = express.Router()
+const db = prisma
+
+planFormRouter.post('/',
+  body('description').isString(),
+  body('startDate').isString(),
+  body('startDate').isString(),
+  body('endDate').isString(),
+  body('EventType').isIn(Object.values(EventType)),
+  body('friends').isArray(),
```
justinpchen94: is this usage correct? this feels like it's creating a tonnnn of validation middlewares

FWIW I've used `express-validator` in conjunction with the `joi` package in the past and this is pretty standard in case you're looking for an alternative

backend/routes/planForm.ts
```diff
@@ -0,0 +1,60 @@
...
+          {
+            userID: friend.id,
+            planID: plan.id,
+            inviteStatus: InviteStatus.INVITED,
+            role: Role.FRIEND
+          }
+        ))
+      })
+    } catch (err) {
+      res.status(400).send('Invalid form submission')
```
justinpchen94: is 400 the right code here? we technically don't know if an error here represents a bad input or some other 500 error right?

justinpchen94 (commented):
some small nits but will wait until frontend is merged to main and you rebase this branch to get rid of the frontend diffs before approving

backend/routes/planForm.ts
kdata12: Added error handling as a todo task

backend/routes/planForm.ts
kdata12: I believe that this is the right usage. Here's the doc: https://express-validator.github.io/docs/#basic-guide

Can you show me what you mean by using `express-validator` with `joi`. Also, I'm new to `express-validator`.

justinpchen94: try checking some examples out like this https://www.npmjs.com/package/express-joi-validation

backend/routes/planForm.ts
kdata12: Can I just keep the error there as a placeholder until I actually implement proper error handling?

backend/app.ts
```diff
@@ -1,175 +1,181 @@
...
+)
 
 app.use(passport.session())
 
 passport.serializeUser((user: any, done: any) => {
-    return done(null, user.id)
+  return done(null, user.id)
 })
 
+// eslint-disable-next-line @typescript-eslint/no-misused-promises
```
justinpchen94: prob just disable this rule

backend/routes/planForm.ts
```diff
@@ -0,0 +1,60 @@
...
+  async (req, res) => {
+    const {
+      description,
+      startDate,
+      endDate,
+      EventType,
+      friends
+    } = req.body
+    try {
+      const plan = await db.plan.create({
```
justinpchen94: so in general we don't usually want a route file like this to contain the code that directly modifies the db. a standard name for this is called a "3 layer service architecture" 

see this article: https://blog.devgenius.io/the-three-layer-architecture-for-node-js-applications-ce32a3a30fa6

the names are relatively interchangeable but typically I see/use "routes" "controllers" and "models" as the main 3 layers. 

In this route file, we receive and validate the request schema and in the request handler we are passed the `req` and `res`. generally we never pass the `req` and `res` further down and leave it in this file. 

we pull out the values we need, and pass them into `controller` methods. in this example i'd expect there to be a `backend/controllers/plan.ts` file of some sort and it would have a method called like `createPlan` that takes in the values in your `req.body`.  generally all of the true business logic lives inside the controller file

And then finally we have `model` files, which would house all the code that directly calls the prisma client and calls the actual db methods. all files wishing to make db updates would then be able to go through methods in this file, rather than having db calls just spread out everywhere

backend/routes/planForm.ts
justinpchen94: as long as you add a todo!

justinpchen94 (approved):
lgtm to unblock but you could do the refactor in a different pr if you'd like

backend/service/planMembership.ts
```diff
@@ -0,0 +1,10 @@
+import PlanMembershipModel from '../models/planMembership';
+
+export default class PlanMembership {
+  public static async initiatePlanMembership (data: any): Promise<any> {
+    const leaderPlanMembership = await PlanMembershipModel.dbCreateOnePlanMembership(data)
+    const friendsPlanMembership = await PlanMembershipModel.dbCreateManyPlanMembership(data)
```
kdata12: `const friendsPlanMembership` is not working at the moment because we haven't set up the frontend to send a the proper array of users that prisma wants

backend/models/plan.ts
```diff
@@ -0,0 +1,24 @@
+import prisma from '../utils/prismaClient'
+import { type Plan } from '@prisma/client'
+const db = prisma
+
+export default class PlanModel {
+  public static async dbCreateOneplan (incomingData: Plan): Promise<any> {
+    try {
+      const responseData = await db.plan.create({
+        data: {
```
kdata12, line 9: For this data, should we be more explicit and readable by typing `incoming.description`, `incoming.startDate` and etc. Or should we do `data: incomingData` here

backend/app.ts
justinpchen94: can we rm this comment if the rule has been turned off?

backend/controllers/plan.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const planData = {
+      description,
+      startDate: new Date(startDate),
+      endDate: new Date(endDate),
+      specialEventType: eventType
+    }
+    const planRecord = await PlanService.initiatePlan(planData)
+
+    if (planRecord === undefined) throw new Error('initiate plan failed')
```
kdata12, line 22: placeholder error handlers

backend/app.ts
```diff
@@ -1,175 +1,180 @@
...
+      done(null, user)
+    }
+  }
+  ))
 
 app.use('/', testRouter)
 app.use('/auth/google', googleOAuthRouter)
 app.use('/auth/facebook', facebookOAuthRouter)
 app.use('/login', loginAuthRouter)
+app.use('/api/planForm', checkUserAuthorization, planFormRouter)
```
justinpchen94: nit: in following with REST naming conventions, this should just be `/api/plan`

backend/controllers/plan.ts
justinpchen94: do you have to check for null here as well or is the expected return `undefined`?

backend/models/plan.ts
justinpchen94: honestly when in doubt, more explicit the better is generally the rule of thumb

backend/models/plan.ts
```diff
@@ -0,0 +1,24 @@
...
+          description: incomingData.description,
+          startDate: incomingData.startDate,
+          endDate: incomingData.endDate,
+          specialEventType: incomingData.specialEventType
+        }
+      })
+
+      if (responseData === undefined) throw new Error(`Plan was not created: ${JSON.stringify(incomingData)}`)
```
justinpchen94, line 17: can just do `if (responseData)` i believe

backend/models/plan.ts
```diff
@@ -0,0 +1,24 @@
...
+
+      return responseData
+    } catch (err) {
+      console.log(`Error in dbCreateOnePlan: ${err}`)
```
justinpchen94: should be a `console.error`

backend/models/planMembership.ts
```diff
@@ -0,0 +1,41 @@
...
+          role: Role.PLAN_LEADER,
+          inviteStatus: InviteStatus.NOT_APPLICABLE,
+          userID: data.userID,
+          planID: data.planID
+        }
+      })
+
+      return responseData
+    } catch (err) {
+      console.log(`Error in db.planMembership.create: ${err}`)
```
justinpchen94: same - `console.error`

backend/models/planMembership.ts
```diff
@@ -0,0 +1,41 @@
...
+            planID: data.planID,
+            inviteStatus: InviteStatus.INVITED,
+            role: Role.FRIEND
+          }
+        ))
+      })
+
+      return responseData
+    } catch (err) {
+      console.log(`Error in db.planMembership.createMany: ${err}`)
```
justinpchen94: `console.error`

backend/modules/auth.ts
```diff
@@ -0,0 +1,9 @@
+import { type NextFunction, type Request, type Response } from 'express'
+
+export const checkUserAuthorization = (req: Request, res: Response, next: NextFunction): void => {
+  if (req.user === null) {
+    res.status(401).json({ error: 'Not Authorized' })
```
justinpchen94: will this handle a redirect?

justinpchen94 (approved):
mostly small nits, looking good!

backend/controllers/plan.ts
kdata12: For the prisma create methods, they either return the data or undefined.

backend/modules/auth.ts
kdata12: yeah i think it should redirect the user to an error page

backend/models/plan.ts
kdata12: this would throw an error when responseData has been fulfilled

backend/modules/auth.ts
kdata12: right now, this is just going to redirect to the main page as a placehodler. Added as a task

PR #44, opened 2023-03-31
T126: Create context for user login

kdata12:
## Description 

Handle User log in states in the frontend

## Types of changes

## Checklist


frontend/src/components/contacts.tsx
```diff
@@ -1,136 +1,138 @@
-import React, { useState } from "react";
-import axios from "axios";
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
justinpchen94: really would recommend not ignoring this one - why would you want to declare variables that you don't ever use?

frontend/src/components/contacts.tsx
```diff
@@ -1,136 +1,138 @@
...
+  }
 
-  const handleRemoveRelationshipType = (index: number) => {
+  const handleRemoveRelationshipType = (index: number): void => {
     setRelationships((prevRelationships) => [
       ...prevRelationships.slice(0, index),
-      ...prevRelationships.slice(index + 1),
-    ]);
-  };
+      ...prevRelationships.slice(index + 1)
```
justinpchen94: is this code meant to simply remove an entry? look into the `splice` method

kdata12: @fahadakhan2

backend/app.ts
```diff
@@ -175,5 +176,6 @@ app.use('/auth/facebook', facebookOAuthRouter)
 app.use('/login', loginAuthRouter)
 app.use('/api/plan', checkUserAuthorization, planFormRouter)
 app.use('/signup', signUpAuth)
+app.use('/user-data', userDataRouter)
```
justinpchen94: let's not mix kebab case with camel case throughout the app

backend/routes/userData.ts
```diff
@@ -0,0 +1,14 @@
+import express from 'express'
+const userDataRouter = express.Router()
+
+userDataRouter.get('/', (req, res) => {
+  console.log('hitting user-data')
+  console.log(req.user)
```
justinpchen94: either convert these into console.info or rm these debug console logs

backend/routes/userData.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (req.user !== null && req.user !== undefined) {
```
justinpchen94, line 5: nit: this can just be `if (req.user) {`

backend/routes/userData.ts
```diff
@@ -0,0 +1,14 @@
+import express from 'express'
+const userDataRouter = express.Router()
+
+userDataRouter.get('/', (req, res) => {
```
justinpchen94: kind of a strange route imo - the full path for this is `wewrap.com/user-data/` which in traditional REST nomenclature would normally be used for a bulk GET endpoint, aka GET me many users. the correct way to structure this endpoint is as `/userData/:id`.

frontend/src/App.tsx
```diff
@@ -1,23 +1,26 @@
...
-      </Routes>
-    </div>
+      <GetUserContext>
+        <Routes>
+          <Route element={<ProtectedRoutes />}>
+            <Route path="/contactPage" element={<Contacts />} />
+          </Route>
+          <Route path="/login" element={<LoginForm />} />
+          <Route path="/signup" element={<SignUp />} />
+          <Route path="/tempLandingPage" element={<TempLandingPage />} />
```
justinpchen94, line 25: did you intentionally remove the /planForm page?

frontend/src/components/UserContext.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { createContext } from 'react'
+import { fetchUser } from '../hooks/fetchUser'
+
+export const UserContext = createContext<any>(null)
+
+export const GetUserContext = ({ children }: any): JSX.Element => {
+  const user = fetchUser()
```
justinpchen94, line 8: shouldn't this be an async function requiring await? i haven't done this specific thing you're doing here so I might be wrong

frontend/src/components/protectedRoutes.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+  if (user !== undefined && user !== null) {
+    return (<Outlet />)
+  } else if (loadingStatus === 'unloaded' || loadingStatus === 'loading') {
+    return (<h1>Loading user data...</h1>)
+  } else if (loadingStatus === 'error') {
+    return (<h1>Error fetching user</h1>)
+  } else {
+    return (<h1>Not authorized</h1>)
+  }
```
justinpchen94: b/c you are `return`ing in every clause, you can omit the `else` and just have this be a series of `if` statements and not `else if`

frontend/src/components/tempLandingPage.tsx
```diff
@@ -1,9 +1,12 @@
 /* eslint-disable @typescript-eslint/no-misused-promises */
 import axios from 'axios'
 import { useNavigate } from 'react-router-dom'
-import React from 'react'
+import React, { useContext } from 'react'
+import { UserContext } from './UserContext'
 
 export const TempLandingPage = (): JSX.Element => {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
justinpchen94: bro why

backend/routes/userData.ts
kdata12: I'm doing it that way cause of the eslint strict boolean expression rule.
image

backend/routes/userData.ts
kdata12: done!

frontend/src/App.tsx
kdata12: Added back

backend/routes/userData.ts
kdata12: From the frontend, we wouldn't have a user id to send to the backend, that's why I did `/userData/`. The route just returns the user from the cookie.

frontend/src/components/UserContext.tsx
kdata12: Yeah i didn't find the need to use an await here cause the custom hook is not returning a promise, plus the useEffect callback fn in `fetchUser()` is taking care of async API request

frontend/src/components/protectedRoutes.tsx
kdata12: ok fixed

frontend/src/components/tempLandingPage.tsx
kdata12: bruhhhh

backend/routes/userData.ts
kdata12: rm'ed them

backend/routes/signUpAuth.ts
```diff
@@ -4,37 +4,36 @@ import prisma from '../utils/prismaClient';
...
+      const user = await db.user.create({
+        data: {
+          email: req.body.email,
+          firstName: req.body.firstName,
+          lastName: req.body.lastName,
+          password: hashedPassword.toString('base64'),
+          salt
+        }
+      })
+      console.log(user); // redirect to the login page
```
justinpchen94: rm extraneous logs

backend/routes/signUpAuth.ts
```diff
@@ -4,37 +4,36 @@ import prisma from '../utils/prismaClient';
...
+    })
+    res.status(200);
```
justinpchen94: hm i think thinking deeper, this res.status should actually be inside the fn above - the crypto fn is callback based so line 20 will run and then immediately skip to line 35. the res.status should only be sent after we confirm the user has been created

backend/routes/userData.ts
justinpchen94: gotcha - explicitness isn't a bad thing, worth keeping it the way you have it then

frontend/src/components/UserContext.tsx
justinpchen94: hmm okay i won't block on this issue but something here feels off - at the very least can you add some more explicit typings here instead of a bunch of `any`?

frontend/src/components/protectedRoutes.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { UserContext } from './UserContext'
+
+export const ProtectedRoutes = (): any => {
+  const [user, loadingStatus] = useContext(UserContext)
+
+  if (user !== undefined && user !== null) {
+    return (<Outlet />)
+  }
+  if (loadingStatus === 'unloaded' || loadingStatus === 'loading') {
+    return (<h1>Loading user data...</h1>)
```
justinpchen94, line 13: will this second and third if statements ever get run? won't the user have a nil value always while loading?

frontend/src/hooks/fetchUser.ts
```diff
@@ -0,0 +1,24 @@
...
+        setLoadingStatus('loading')
+        const userDataResponse = await axios.get('http://localhost:8000/userData/1', { withCredentials: true })
+        setUser(userDataResponse.data)
+        setLoadingStatus('loaded')
+      } catch (error) {
+        setLoadingStatus('error')
+        console.error(error)
+      }
+    }
+    void getUserData()
```
justinpchen94, line 20: does this need to even get wrapped in a fn? what benefit do we get in defining the fn and then calling it here rather than calling this code directly or pulling it out into a separate fn outside this component?

justinpchen94: i don't know if the catch in the getUserData gets thrown correctly if this line 20 here doesn't also include a .catch block

backend/routes/signUpAuth.ts
kdata12: This make sense. Resolved

frontend/src/components/UserContext.tsx
kdata12: added explicit typings

frontend/src/components/protectedRoutes.tsx
kdata12: the 2nd and 3rd if statements worked when I tested them. the 2nd statement is the same as `user === null` while loading, so you're correct

frontend/src/hooks/fetchUser.ts
kdata12: yes, the catch block caught the errors when tested it. 

perhaps this was something you're looking for?

image

frontend/src/components/contacts.tsx
```diff
@@ -24,7 +24,7 @@ export const Contacts = (): JSX.Element => {
   const [phoneNumber, setPhoneNumber] = useState<string>('')
   const [notes, setNotes] = useState<string>('')
   const [errorMessage, setErrorMessage] = useState<string>('')
-  const [showError, setShowError] = useState<boolean>(false)
+  const [, setShowError] = useState<boolean>(false)
```
kdata12: fahad had this fixed in his branch already

backend/routes/signUpAuth.ts
```diff
@@ -4,37 +4,33 @@ import prisma from '../utils/prismaClient';
...
-                    lastName: req.body.lastName, 
-                    password: hashedPassword.toString("base64"),
-                    salt 
-                },
-            })
-            console.log(user); //redirect to the login page 
-        })
-        res.status(200);
+signUpAuthRouter.post('/', async function (req, res, next) {
+  const existingUser = await db.user.findUnique({
```
justinpchen94, line 8: not blocking but this would be better off abstracted into a method in a `model` file since it is directly interacting with the db and we shouldn't do that directly in the route hander

justinpchen94 (approved):

backend/routes/signUpAuth.ts
kdata12: added as a todo!

PR #46, opened 2023-04-17
T142: Create database seed with a test user and friends

kdata12:
## Description 

Created a database seed that will insert one test user and friend relationships with other dummy users. 

## Types of changes

## Checklist

- [x] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in downstream modules


backend/prisma/userSeed.ts
```diff
@@ -0,0 +1,141 @@
...
+          salt
+        }
+      })
+    })
+  } catch (err) {
+    console.error(err)
+  }
+
+  // Execute 1 sec later
+  setTimeout(async () => {
```
kdata12: So I discovered that the `crypto.pbkdf2` fn takes some time to create a hash value, and afterward, create a test user record. This setTimeout solution guarantees that a test user record was created before moving on to the next step.

backend/prisma/userSeed.ts
```diff
@@ -0,0 +1,141 @@
...
+const testUser = {
+  email: 'test@gmail.com',
+  firstName: 'bob',
+  lastName: 'admin'
+}
+
+const main = async (): Promise<void> => {
+  // create a test user,
+  try {
+    const salt = crypto.randomBytes(16).toString('base64');
```
justinpchen94: i won't really block this pr since this is just a db seed script but consider how you could've just written the original production fn in a way that is exportable in a fn such as `createUserWithHashedPw` and then just called that here instead, rather than copy pasting the logic

backend/prisma/userSeed.ts
justinpchen94: while this is probably fine given this is a seed script, in reality you are mixing Callback style js programming with async/await. due to the inherently async nature of js, your line 64 begins an asynchronous execution of code but does not block the rest of this script from running. the best thing you could do here is use a promisification library such as bluebird to asyncify the crypto.pbkdf2 call and convert the callback into a promise. alternatively I think you could actually do it without al library but this presents an opportunity to look into those libraries if you wish to

backend/prisma/userSeed.ts
```diff
@@ -0,0 +1,141 @@
...
+        firstName: testUser.firstName
+      }
+    })
+    // create many users
+    await db.user.createMany({
+      data: realUsers
+    })
+
+    // bob adds a bunch of user
+    realUsers.map(async user => {
```
justinpchen94: part of the problem here too is that the .map call is not asynchronous, so it is not going to pause execution of js on this entire .map call finishing. it will queue up all of these async fn's and they will just kind of finish whenever they happen to finish. you could try looking into using Promise.all here

backend/prisma/userSeed.ts
```diff
@@ -0,0 +1,141 @@
...
+      })
+    })
+  }, 1000)
+}
+
+main()
+  .then(async () => {
+    console.log('Script executed successfully')
+    await db.$disconnect()
+    console.log('Disconnected fom DB')
```
justinpchen94: typo

backend/prisma/userSeed.ts
kdata12: i will keep this in mind next time I copy and paste code

backend/prisma/userSeed.ts
kdata12: Got this taken care of by promisifying the cypto function

backend/prisma/userSeed.ts
kdata12: Fixed this with promise.all()

backend/prisma/userSeed.ts
justinpchen94: i love it, good shit!

backend/prisma/userSeed.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const testUser = {
+  email: 'test@gmail.com',
+  firstName: 'bob',
+  lastName: 'admin'
+}
+
+const main = async (): Promise<void> => {
+  // create a test user
+  const cryptoPromise = (): unknown => {
```
justinpchen94: nit: it's not unknown, it should be a `User` right?

justinpchen94 (approved):
lgtm pending 1 comment

backend/prisma/userSeed.ts
kdata12: fixed! with `promise<User>`

PR #48, opened 2023-04-24
T154: Add member gift idea data model

kdata12:
## Description 

Add data model for member gift ideas

image

## Types of changes

## Checklist


justinpchen94 (commented):
i think you need to still generate the migration files with this change right? or are you just requesting a review on schema

kdata12:
> i think you need to still generate the migration files with this change right? or are you just requesting a review on schema

Yeah I'm requesting a review on schema

PR #59, opened 2023-05-17
T123: Send email to contacts email backend

kdata12:
## Description 

Create a route that sucessfully send out plan invitations to the selected contacts email.

One way that I handle 'spamming invites' from the frontend is by only allowing users to reinvite their friends after the invite expiration (2 days) date has expired. This is a temp fix, but I think a better fix would be only loading contacts that were not invited in the frontend instead of loading all contacts, regardless if they were invited or not invited.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have made sure that eslint was passed locally


kdata12:
You can ignore all of the frontend files, I created this branch off of hyun/plan-search-bar branch so that I can quickly test my api.

backend/awsSes/SesService.ts
```diff
@@ -0,0 +1,54 @@
+import { type User } from '@prisma/client';
+import AWS from 'aws-sdk'
+
+AWS.config.update({ region: 'us-west-1' });
+export default class SesService {
+  public static async sendMail(planLeader: User, url: string, email: string, message: string) {
+    const baseURL = 'http://localhost:3000'
```
chinkevin25: I think this eventually will need to become a ENV var so when we deploy it won't point to localhost

backend/awsSes/SesService.ts
```diff
@@ -0,0 +1,54 @@
...
+    const planLeaderFullName = `${planLeader.firstName} ${planLeader.lastName}`
+    const params = {
+      Destination: {
+        ToAddresses: [
+          `${email}`
```
chinkevin25: nit this is unnecessary string templating, your email is already a string so you can just leave this as `email`

backend/awsSes/SesService.ts
```diff
@@ -0,0 +1,54 @@
...
+              <a href=${baseURL}${url}>
+                Join Plan
+              </a>
+            </button>
+            `
+          }
+        },
+        Subject: {
+          Charset: 'UTF-8',
+          Data: `${planLeaderFullName} has sent you an invite to a Wewrap plan!`
```
chinkevin25: nit: consistency w/ branding would be `WeWrap` right?

backend/controllers/inviteContacts.ts
```diff
@@ -0,0 +1,37 @@
...
+  public static async inviteContacts(req: any, res: any, next: NextFunction): Promise<void> {
+    try {
+      const {
+        message,
+        selectedContacts: contacts,
+        planID
+      } = req.body
+
+      const inviteContactsRes = await InviteContactService.setupEmailInviteToContacts(req.user, planID, contacts, message)
+
```
chinkevin25: `setupEmailinviteToContacts` has the potential to `throw` do we need to wrap this in a `try catch` block?

backend/controllers/inviteContacts.ts
```diff
@@ -0,0 +1,37 @@
...
+      if (inviteContactsRes === null) throw new Error('Unable to send invites to all contacts')
+
+      if (inviteContactsRes.length !== 0) {
+        console.log({
+          contactsNotInvited: inviteContactsRes
+        })
+        res.status(201).json({
```
chinkevin25: nit: is 201 correct? 201 is "created" right, but in this specific case it's not a 201 I think. Not really important in this context, but just worth thinking about

backend/controllers/inviteContacts.ts
```diff
@@ -0,0 +1,37 @@
+import { type NextFunction } from 'express';
+import InviteContactService from '../service/inviteContact';
+
+export default class InviteContactController {
+  public static async inviteContacts(req: any, res: any, next: NextFunction): Promise<void> {
```
chinkevin25: if we're being pedantic, these types are `Express.Request` and `Express.Response` I think?

backend/models/planInvite.ts
```diff
@@ -0,0 +1,65 @@
...
+const db = prisma
+
+export default class PlanInviteModel {
+  private static getNthDateFromToday(numOfDay: number): Date {
+    const today = new Date()
+    const nthDayFromToday = new Date(today)
+    nthDayFromToday.setDate(nthDayFromToday.getDate() + numOfDay)
+
+    return nthDayFromToday
+  }
```
chinkevin25: This seems like a nice utility function! It's okay to leave it here if you'd like, but a common thing is to create a utility folder / module so other people can use this function.

backend/models/planInvite.ts
```diff
@@ -0,0 +1,65 @@
...
+          expiration: PlanInviteModel.getNthDateFromToday(1)
+        }
+      })
+
+      return responseData
+    } catch (error) {
+      console.error(`Error in dbCreateOnePlanInvite: ${error}`)
+      return null
+    }
+  }
```
chinkevin25: small nit: you can make the `email` arg `inviteeEmail`, which would let you do the short hand object syntax and also be more specific that the email to provide is the `invitee` rather than the `inviter`. Small readability optimization

backend/modules/auth.ts
```diff
@@ -1,9 +1,10 @@
 import { type NextFunction, type Request, type Response } from 'express'
 
 export const checkUserAuthorization = (req: Request, res: Response, next: NextFunction): void => {
-  if (req.user === null) {
+  if (req.user === null || req.user === undefined) {
     res.status(401).json({ error: 'Not Authorized' }).redirect('http://localhost:3000/')
   } else {
+    console.log('auth successfull')
```
chinkevin25: is this a debug log that we wanted to keep or remove?

backend/prisma/schema.prisma
```diff
@@ -153,3 +154,13 @@ model PlanMembership {
   inviteStatus InviteStatus
   role         Role
 }
+
+model PlanInvite {
+  id        String   @id @default(uuid())
+  plan      Plan     @relation(fields: [planID], references: [id])
+  planID    String
+  inviteeEmail String
```
chinkevin25, line 163: I've been abset in a few convos. the Invitee Does't need to have a wewap account right? So we don't need to make a FK relation to the User's table here?

backend/routes/inviteContactsRouter.ts
```diff
@@ -0,0 +1,22 @@
...
+inviteContactsRouter.post('/',
+  body('planID').isString().notEmpty(),
+  body('message').isString().notEmpty(),
+  body('selectedContacts')
+    .custom((value: any[]) => {
+      return value.every((contact) => {
+        return 'id' in contact &&
+          'firstName' in contact &&
+          'ownerID' in contact
+      })
```
chinkevin25: this is fine, but you can also use `Object.hasOwn()` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

backend/service/inviteContact.ts
```diff
@@ -0,0 +1,94 @@
...
+
+            await InviteContactService.sendEmailInviteToContact(planLeader, newUrl, contactEmail, message)
+
+            break;
+          case PlanInviteStatus.HAS_INVITED:
+
+            alreadyInvitedContacts.push(contact)
+            break;
+        }
+      }))
```
chinkevin25: the way this block is structure implies something different.

You're calling `Promise.all` so we'd assume you'd end up with an Array of Promises. You're using a `map` which implies that you're convert one array to converting it to another. 

but as you'll notice in your block -- youre not actually returning any promises, in fact you're returning undefined in all cases, so you have a `Promise<undefined>[]` return type right now. 

Also related, using HOF ,like Map andForEach, with Promises has some funky behavior. https://stackoverflow.com/questions/64978604/async-await-map-not-awaiting-async-function-to-complete-inside-map-function-befo#:~:text=map()%20is%20not%20async,not%20waiting%20for%20that%20promise.

TLDR: maps are not promise aware, so it can lead to some un-expected results. I don't think you do in this case because of the Promise.all, but just wanted to share.

Also I'm just kind of confused what this function does. why does this function return the list of "emails already invited" rather than the list of all emails invited?

chinkevin25: Also, in general, using a `push` inside of a `.map` usually implies that you're doing something a little smelly.

backend/service/inviteContact.ts
```diff
@@ -0,0 +1,94 @@
...
+      return alreadyInvitedContacts
+    } catch (error: any) {
+      console.error(error)
+      console.error(error.stack)
+      return null
+    }
+  }
+
+  private static async checkPlanInviteStatus(planID: string, email: string) {
+    if (await PlanInviteModel.readOnePlanInvite(planID, email) === null) {
```
chinkevin25: nit: To improve readability, I would recommend using these interim variables. Readability > slight space optimizations.

```js
const plan = PlanInviteModel.readOnePlanInvite(plainID, email) 
if (plan === null) {
    // do something
}
```

backend/service/inviteContact.ts
```diff
@@ -0,0 +1,94 @@
...
+      return PlanInviteStatus.NOT_INVITED
+    }
+
+    if ((await InviteContactService.isPlanInviteExpired(planID, email))) {
+      return PlanInviteStatus.INVITE_EXPIRED
+    }
```
chinkevin25: For example, by not having this interim variable, you're actually doing two DB calls instead of 1 which is probably worse for performance (latency vs space). Instead, we should be using the interim variable `plan` to determine if it's expired.

backend/service/inviteContact.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  private static async isPlanInviteExpired(planID: string, email: string): Promise<boolean> {
+    const today = new Date()
+
+    const planInviteRes = await PlanInviteModel.readOnePlanInvite(planID, email);
+
+    if (planInviteRes === null) throw new Error('Unable to find plan invite')
+
+    return planInviteRes.expiration < today
+  }
```
chinkevin25: In general, warning bells should be shooting in your head when you have functions that call `new Date()`. These types of functions are typically harder to test. They are usually harder to test because you have a constantly changing variable that you can't really set. This concept is called "Dependency Injection"

The way I might write this instead, is to create a function that just takes in two dates (the expiration date, and the date to compare against) as arguments.

You then create a more reusable block of code, and you still get the opportunity to use it for your plan logic.

backend/utils/testQuery.ts
chinkevin25, line 1: What is this file for?

chinkevin25 (changes requested):
I think we're also missing changes to your env-sample. How are your AWS secrets configured right now?

backend/service/inviteContact.ts
```diff
@@ -0,0 +1,94 @@
...
+
+        switch (await InviteContactService.checkPlanInviteStatus(planID, contactEmail)) {
+          case PlanInviteStatus.NOT_INVITED:
+            const planInviteResponse = await PlanInviteModel.dbCreateOnePlanInvite(planID, contactEmail);
+
+            if (planInviteResponse === null) throw new Error('planInviteResponse is null')
+
+            const url = InviteContactService.generateInviteLink(planInviteResponse.id);
+
+            await InviteContactService.sendEmailInviteToContact(planLeader, url, contactEmail, message)
```
kdata12, line 28: return

backend/service/inviteContact.ts
```diff
@@ -0,0 +1,94 @@
...
+
+            if (expiredPlanInviteResponse === null) throw new Error('expiredPlanInviteResponse is null')
+
+            const updatePlanInviteRecord = await PlanInviteModel.updatePlanInviteExpiration(expiredPlanInviteResponse.id)
+
+            if (updatePlanInviteRecord === null) throw new Error('modelResponse is null')
+
+            const newUrl = InviteContactService.generateInviteLink(updatePlanInviteRecord.id);
+
+            await InviteContactService.sendEmailInviteToContact(planLeader, newUrl, contactEmail, message)
```
kdata12, line 46: return

backend/service/inviteContact.ts
```diff
@@ -0,0 +1,94 @@
...
+import PlanInviteModel from '../models/planInvite';
+
+enum PlanInviteStatus {
+  HAS_INVITED,
+  NOT_INVITED,
+  INVITE_EXPIRED
+}
+
+export default class InviteContactService {
+  static async setupEmailInviteToContacts(planLeader: User, planID: string, contactsArray: Contact[], message: string): Promise<Contact[] | null> {
```
kdata12: rename sendEmailInvitesToContacts or something better, This function is doing too much stuff.

backend/awsSes/SesService.ts
kdata12: Yes, I will make a pr for this fix soon.

backend/awsSes/SesService.ts
kdata12: LMAOOO

backend/controllers/inviteContacts.ts
kdata12: i changed it to 200 ok

backend/controllers/inviteContacts.ts
kdata12: Ok, I added the Express.Response, but Express.Request is giving me squiggly lines when I try to access `req.body`

kdata12: Ok nvm, it got it to work

backend/models/planInvite.ts
```diff
@@ -45,16 +45,30 @@ export default class PlanInviteModel {
...
+            OR: {
+              ...params
+            }
+          }
+        })
+
+        if (responseData === null) throw new Error(`Unable to find a matching plan invite with in the database${JSON.stringify(params)}`)
+
+        return responseData;
+      }
```
chinkevin25: The simplification you can make is to take advantage of having "AND" be the default.

so you could remove all of this by simply stating

```
where: {
   [options]: {
      ...params
    }
}
```
and then default your `options` to `AND`
```  public static async readOnePlanInvite(params: PlanInviteModelReadInput, options?: 'AND' | 'OR' = 'AND'): Promise<PlanInvite | null> {```

or whatever the syntax is i always forget w/ typescript

File: backend/tsconfig.json [hunk 20]
```diff
@@ -10,7 +10,7 @@
         // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
         /* Language and Environment */
         "target": "es2017", /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
-        // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
+        "lib": ["es2022"],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
```
chinkevin25, line 13: why this change?

chinkevin25 (approved):

backend/tsconfig.json
kdata12: Typescript was throwing an error when I tried to use `Object.hasOwn()`. The suggested fix was putting this here.

PR #65, opened 2023-05-30
T174: Redirect user after clicking on plan invite link

kdata12:
## Description 

Feature: After clicking o the plan invite link in their email, the user will be redirect to a 'Plan invite redirect' page which will redirect the user based on their cookie information. They will redirect to the log in page or unauthorized access page or plan page

https://www.loom.com/share/459da77df4454ddcac5a1e792ab37bee

## Types of changes

## Checklist


backend/controllers/inviteContacts.ts
```diff
@@ -35,4 +36,57 @@ export default class InviteContactController {
...
+      const planInviteID = req.params.id
+
+      // if no cookie, redirect to log in
+      if (req.user === null || req.user === undefined) {
+        res.status(200).json({
+          reason: 'NOT_LOGGED_IN',
+          planInviteID
+        })
+        return
+      }
```
chinkevin25: Hmm, should this be done here, or is this not handled already in your route middleware?

also, if we're going to return a status code here, it would be 401

backend/controllers/inviteContacts.ts
```diff
@@ -35,4 +36,57 @@ export default class InviteContactController {
...
+        res.status(200).json({
+          reason: 'USER_ALREADY_A_PLAN_MEMBER',
+          planID
+        })
+        return
+      }
+
+      // cookie is present, email match, set up plan membership
+      const planID = await PlanInviteService.setUpInviteePlanMembership(planInviteID, req.user)
+      console.log('email match')
```
chinkevin25: rm

backend/routes/verifyPlanInvite.ts
```diff
@@ -0,0 +1,7 @@
+import express from 'express'
+import InviteContactController from '../controllers/inviteContacts';
+const verifyPlanInviteRouter = express.Router()
+
+verifyPlanInviteRouter.get('/:id', InviteContactController.verifyPlanInvite)
```
chinkevin25, line 5: I am thinking your auth logic could go as a middle ware here

backend/service/planInvite.ts
```diff
@@ -0,0 +1,70 @@
...
+    })
+
+    if (planInviteData === null) throw new Error('unable to find plan')
+
+    const isUserInPlan = await PlanMembershipModel.dbReadOnePlanMembership({ planID: planInviteData.planID, userID })
+
+    // if no record is found, then user is not in plan
+    if (isUserInPlan === null) return false
+
+    return true
```
chinkevin25: maybe less readable, so no change needed, but i'd say we could `return !!isUserInPlan`

backend/controllers/inviteContacts.ts
```diff
@@ -35,4 +36,57 @@ export default class InviteContactController {
...
+        res.status(200).json({
+          reason: 'NOT_LOGGED_IN',
+          planInviteID
+        })
+        return
+      }
+
+      const isEmailMatch = await PlanInviteService.isUserEmailMatchPlanInviteEmail(req?.user?.email, planInviteID)
+
+      if (isEmailMatch === null) throw new Error('fail to run isUserEmailMatchPlanInviteEmail')
```
chinkevin25: Do we need this error? If the email check doesnt run, and returns null is that any different than just email and logged in acc not matching?

backend/service/planInvite.ts
```diff
@@ -0,0 +1,70 @@
...
+      // search the PlanInvite model to see if there is a record of the inviteeEmail and planInviteID
+      const res = await PlanInviteModel.readOnePlanInvite({
+        inviteeEmail,
+        id: planInviteID
+      })
+
+      // if no record is found, then email does not match
+      if (res === null) return false
+
+      return true
```
chinkevin25: kind of the same here, we could just return `!!res` provided the readOnePLanInvite returns undefined/null if it cannot find .

Also REALLY minor point, but technically instead of doing a readOne (which returns a full document), you could just return the count instead

https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#count

frontend/src/index.tsx
```diff
@@ -7,9 +7,7 @@ const root = ReactDOM.createRoot(
   document.getElementById('root') as HTMLElement
 )
 root.render(
-  <React.StrictMode>
```
chinkevin25: any reason were removing strict mode?

chinkevin25 (changes requested):

backend/service/planInvite.ts
kdata12: This is the first time i've seen this syntax

backend/controllers/inviteContacts.ts
kdata12: That's very true..if `isEmailMatch` is null then the user's email did not match too

backend/service/planInvite.ts
kdata12: gotcha

frontend/src/index.tsx
kdata12: Oh i thought strict mode would do 2 POST request and which would send 2 emails to the user.

backend/routes/verifyPlanInvite.ts
kdata12: yup i will put the `checkUserAuthorization` middleware at the router level

backend/controllers/inviteContacts.ts
kdata12: Rm'ed this

kdata12:
Tested the changes, everything is working as expected.

chinkevin25 (approved):

